PR #25, opened 2019-09-12
Aws iam

yuval-k:
draft till i add tests.
here for CI

### DIFF
--- a/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
@@ -23,10 +23,17 @@ namespace Extensions {
 namespace HttpFilters {
 namespace AwsLambda {
 
+namespace {
 struct RcDetailsValues {
   const std::string FunctionNotFound = "aws_lambda_function_not_found";
+  const std::string FunctionNotFoundBody =
+      "no function present for AWS upstream";
+  const std::string CredentialsNotFound = "aws_lambda_credentials_not_found";
+  const std::string CredentialsNotFoundBody =
+      "no credentials present for AWS upstream";
 };
 typedef ConstSingleton<RcDetailsValues> RcDetails;
+} // namespace
 
 class AWSLambdaHeaderValues {
 public:
@@ -47,8 +54,10 @@ const HeaderList AWSLambdaFilter::HeadersToSign =
          Http::Headers::get().ContentType});
 
 AWSLambdaFilter::AWSLambdaFilter(Upstream::ClusterManager &cluster_manager,
-                                 TimeSource &time_source)
-    : aws_authenticator_(time_source), cluster_manager_(cluster_manager) {}
+                                 TimeSource &time_source,
+                                 AWSLambdaConfigConstSharedPtr filter_config)
+    : aws_authenticator_(time_source), cluster_manager_(cluster_manager),
+      filter_config_(filter_config) {}
 
 AWSLambdaFilter::~AWSLambdaFilter() {}
 
@@ -59,8 +68,32 @@ AWSLambdaFilter::decodeHeaders(Http::HeaderMap &headers, bool end_stream) {
       const AWSLambdaProtocolExtensionConfig>(
       SoloHttpFilterNames::get().AwsLambda, decoder_callbacks_,
       cluster_manager_);
+
   if (!protocol_options_) {
     return Http::FilterHeadersStatus::Continue;
+  }
+
+  const std::string *access_key{};
+  const std::string *secret_key{};
+  if (protocol_options_->accessKey().has_value() &&
+      protocol_options_->secretKey().has_value()) {
+    access_key = &protocol_options_->accessKey().value();
+    secret_key = &protocol_options_->secretKey().value();
+  } else {
+    auto credentials = filter_config_->getCredentials();
+    if (credentials) {
+      if (!credentials->accessKeyId() || !credentials->secretAccessKey()) {
+        access_key = &credentials->accessKeyId().value();
+        secret_key = &credentials->secretAccessKey().value();
+      }
+    }
+  }
+
+  if ((access_key == nullptr) || (secret_key == nullptr)) {
+    decoder_callbacks_->sendLocalReply(
+        Http::Code::NotFound, RcDetails::get().CredentialsNotFoundBody, nullptr,
+        absl::nullopt, RcDetails::get().CredentialsNotFound);
+    return Http::FilterHeadersStatus::StopIteration;
   }
 
   route_ = decoder_callbacks_->route();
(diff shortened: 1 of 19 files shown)


source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -47,33 +54,52 @@ const HeaderList AWSLambdaFilter::HeadersToSign =
...
+      if (maybeAccessKeyId.has_value() && maybeSecretAccessKey.has_value()) {
+        access_key = &maybeAccessKeyId.value();
+        secret_key = &maybeSecretAccessKey.value();
+      }
+    }
+  }
+
+  if ((access_key == nullptr) || (secret_key == nullptr)) {
+    decoder_callbacks_->sendLocalReply(
+        Http::Code::NotFound, RcDetails::get().CredentialsNotFoundBody, nullptr,
```
EItanya: Nit: is this really a 404: seems more like a 403 unauthorized

source/extensions/filters/http/aws_lambda/config.cc
```diff
@@ -1,21 +1,139 @@
...
+// minutes. refresh itself will only happen when deemed needed by the AWS code.
+//
+// According to the AWS docs
+// (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html)
+//
+// >>> it should get a refreshed set of credentials every hour, or at least 15
+// >>> minutes before the current set expires.
+//
+// Refreshing every 14 minutes should guarantee us fresh credentials.
+constexpr std::chrono::milliseconds REFRESH_AWS_CREDS =
```
EItanya: make this value configurable?

source/extensions/filters/http/aws_lambda/config.cc
```diff
@@ -1,21 +1,139 @@
...
+
+  // tls_slot_ != nil IFF provider_ != nil
+  return tls_slot_->getTyped<ThreadLocalState>().credentials_;
+}
+
+void AWSLambdaConfigImpl::timerCallback() {
+  // get new credentials.
+  auto new_creds = provider_->getCredentials();
+  if (new_creds == CommonAws::Credentials()) {
+    ENVOY_LOG(warn, "can't get AWS credentials - credentials will not be "
```
EItanya: No stream log? is that because this is in the filter config, and therefore out of the request thread?

source/extensions/filters/http/aws_lambda/config.cc
```diff
@@ -1,21 +1,139 @@
...
+                    "refreshed and request to AWS may fail");
+    // TODO: add a stat here
+  } else {
+    // TODO: add a stat here
+    auto currentCreds = getCredentials();
+    if (currentCreds == nullptr || !((*currentCreds) == new_creds)) {
+      ENVOY_LOG(debug, "refreshing AWS credentials");
```
EItanya: same as above?

source/extensions/filters/http/aws_lambda/config.cc
```diff
@@ -1,21 +1,139 @@
...
+  return tls_slot_->getTyped<ThreadLocalState>().credentials_;
+}
+
+void AWSLambdaConfigImpl::timerCallback() {
+  // get new credentials.
+  auto new_creds = provider_->getCredentials();
+  if (new_creds == CommonAws::Credentials()) {
+    ENVOY_LOG(warn, "can't get AWS credentials - credentials will not be "
+                    "refreshed and request to AWS may fail");
+    // TODO: add a stat here
```
EItanya: Any good reason not to add the stat now?

test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc
```diff
@@ -88,6 +109,36 @@ TEST_F(AWSLambdaFilterTest, SingsOnHeadersEndStream) {
   EXPECT_TRUE(headers.has("Authorization"));
 }
 
+TEST_F(AWSLambdaFilterTest, SingsOnHeadersEndStreamWithConfig) {
```
EItanya: `SignsOnHeadersEndStreamWithConfig`

test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc
```diff
@@ -88,6 +109,36 @@ TEST_F(AWSLambdaFilterTest, SingsOnHeadersEndStream) {
...
+                                  {":path", "/getsomething"}};
+  EXPECT_EQ(Http::FilterHeadersStatus::Continue,
+            filter_->decodeHeaders(headers, true));
+
+  EXPECT_TRUE(filter_config_->called_);
+  // Check aws headers.
+  EXPECT_TRUE(headers.has("Authorization"));
+}
+
+TEST_F(AWSLambdaFilterTest, SingsOnHeadersEndStreamWithBadConfig) {
```
EItanya: Signs

test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc
```diff
@@ -88,6 +109,36 @@ TEST_F(AWSLambdaFilterTest, SingsOnHeadersEndStream) {
...
+                                  {":path", "/getsomething"}};
+  EXPECT_EQ(Http::FilterHeadersStatus::StopIteration,
+            filter_->decodeHeaders(headers, true));
+
+  // Check no aws headers.
+  EXPECT_TRUE(filter_config_->called_);
+  EXPECT_FALSE(headers.has("Authorization"));
+}
+
 TEST_F(AWSLambdaFilterTest, SingsOnDataEndStream) {
```
EItanya: Signs

source/extensions/filters/http/aws_lambda/config.cc
yuval-k: not sure it would be useful to make this configurable, as the aws code is hard coded.
we can discuss.

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
yuval-k: i wouldn't say a 403; maybe 400 bad request?

source/extensions/filters/http/aws_lambda/config.cc
yuval-k: correct; the config processing happens in the main thread. there is no stream in this context.

source/extensions/filters/http/aws_lambda/config.cc
yuval-k: yes

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
yuval-k: actually i think the best is 500 - internal server error

PR #27, opened 2019-09-25
Transform extend

EItanya:


source/extensions/filters/http/transformation/transformation_filter.cc
```diff
@@ -65,6 +66,7 @@ Http::FilterDataStatus TransformationFilter::decodeData(Buffer::Instance &data,
   }
 
   if (end_stream) {
+    filter_config_->stats().request_body_transformations_.inc();
```
yuval-k: probably want this line in decodeTrailers as well

source/extensions/filters/http/transformation/transformer.h
```diff
@@ -1,13 +1,40 @@
...
+ * All stats for the transformation filter. @see stats_macros.h
+ */
+#define ALL_TRANSFORMATION_FILTER_STATS(COUNTER, GAUGE)                                                     \
+  COUNTER(request_body_transformations)                                                                         \
+  COUNTER(request_header_transformations)                                                                         \
+  COUNTER(response_header_transformations)                                                                         \
+  COUNTER(response_body_transformations)                                                                         \
+  COUNTER(request_error)                                                                         \
+  COUNTER(response_error)                                                                    \
+  GAUGE(active_faults, Accumulate)
```
yuval-k: is this used?

EItanya: nope, copy pasta

source/extensions/filters/http/transformation/transformation_filter.cc
EItanya: Good call, will add

source/extensions/filters/http/transformation/transformation_filter.h
```diff
@@ -112,6 +113,9 @@ class TransformationFilter : public Http::StreamFilter,
   Http::Code error_code_;
   std::string error_messgae_;
   bool should_clear_cache_{};
+  
+  RouteFilterConfigConstSharedPtr route_filter_config_;
```
yuval-k: i route_filter_config_ is not used - we can remove this?

source/extensions/filters/http/transformation/transformation_filter_config_factory.cc
```diff
@@ -31,12 +37,9 @@ TransformationFilterConfigFactory::createEmptyRouteConfigProto() {
 }
```
yuval-k: i think you can remove createEmptyRouteConfigProto now

source/extensions/filters/http/transformation/transformation_filter_config_factory.h
```diff
@@ -15,21 +16,24 @@ namespace Extensions {
...
+using Extensions::HttpFilters::Common::FactoryBase;
 
-class TransformationFilterConfigFactory : public EmptyHttpFilterConfig {
+class TransformationFilterConfigFactory : public Common::FactoryBase<TransformationConfigProto, RouteTransformationConfigProto> {
 public:
   TransformationFilterConfigFactory()
-      : EmptyHttpFilterConfig(SoloHttpFilterNames::get().Transformation) {}
+      : FactoryBase(SoloHttpFilterNames::get().Transformation) {}
 
   ProtobufTypes::MessagePtr createEmptyRouteConfigProto() override;
```
yuval-k: i think this can be removed

test/extensions/filters/http/transformation/transformation_filter_test.cc
```diff
@@ -43,34 +50,122 @@ class TransformationFilterTest : public testing::Test {
...
     initFilter(); // Re-load config.
   }
 
-  void initFilterWithHeadersBody() {
+  void transformsOnHeaders(TransformationFilterTest::ConfigType configType, unsigned int val) {
+    initFilterWithBodyTemplate(configType, "solo");
 
-    auto &transformation = (*route_config_.mutable_request_transformation());
-    transformation.mutable_header_body_transform();
+    EXPECT_CALL(filter_callbacks_, addDecodedData(_, false)).Times(1);
```
yuval-k: i'm not sure if EXPECT works well in a function that is not a test... but this might have changed since last time i checked

test/extensions/filters/http/transformation/transformation_filter_test.cc
```diff
@@ -43,34 +50,122 @@ class TransformationFilterTest : public testing::Test {
...
+  void happyPathWithBodyPassthrough(TransformationFilterTest::ConfigType configType) {
+    initFilterWithBodyPassthrough(configType);
+
+    auto resheaders = filter_->decodeHeaders(headers_, false);
+    EXPECT_EQ(Http::FilterHeadersStatus::Continue, resheaders);
+
+    Buffer::OwnedImpl downstream_body("{\"a\":\"b\"}");
+    auto res = filter_->decodeData(downstream_body, true);
+    EXPECT_EQ(Http::FilterDataStatus::Continue, res);
+    EXPECT_EQ(0U, config_->stats().request_body_transformations_.value());
```
yuval-k: should the header stats be not zero here? can we test that?

test/extensions/filters/http/transformation/transformation_filter_test.cc
EItanya: This is called only from within a test, I believe that’s why it works

test/extensions/filters/http/transformation/transformation_filter_test.cc
EItanya: Sure

PR #35, opened 2019-10-18
Regex body

yuval-k:


source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -37,72 +42,89 @@ const Http::HeaderEntry *getHeader(const Http::HeaderMap &header_map,
...
-    for (unsigned i = 0; i < group_; i++) {
-      std::advance(submatch_it, 1);
-      if (submatch_it == regex_result.end()) {
-        return "";
-      }
+  std::match_results<absl::string_view::const_iterator> regex_result;
+  if (std::regex_match(value.begin(), value.end(), regex_result,
+                       extract_regex_)) {
+    if (group_ >= regex_result.size()) {
+      // TODO: this indicates user misconfiguration... log or stat?
```
EItanya: Why not log about this now? Or possibly check the number of matches in the constructor and panic there so can check the nack

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -138,56 +176,87 @@ InjaTransformer::InjaTransformer(
...
     break;
-  case envoy::api::v2::filter::http::TransformationTemplate::
-      BODY_TRANSFORMATION_NOT_SET: {
+  case TransformationTemplate::BODY_TRANSFORMATION_NOT_SET: {
     break;
   }
   }
+
+  // parse environment
+    for (char **env = environ; *env != 0; env++) {
```
EItanya: Should this maybe be optional? Not sure how much that would save, but this code is totally unecessary 99% of the time

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -207,6 +276,15 @@ void InjaTransformer::transform(Http::HeaderMap &header_map,
     replace_body(output);
   }
 
+  // DynamicMetadata transform:
+  for (const auto &templated_dynamic_metadata : dynamic_metadata_) {
```
EItanya: Potentially most work regarding dynamic metadata could be done in the log callback because we are not actually planning on using it for any other purpose than logging

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -352,6 +398,133 @@ TEST(InjaTransformer, RemoveHeadersUsingEmptyTemplate) {
...
+
+  auto dynamic_meta = transformation.add_dynamic_metadata_values();
+  dynamic_meta->set_key("foo");
+  dynamic_meta->mutable_value()->set_text("{{body()}}");
+
+  InjaTransformer transformer(transformation);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> filter_callbacks_{};
+
+  EXPECT_CALL(filter_callbacks_.stream_info_, setDynamicMetadata("io.solo.transformation", _)).Times(1);
```
EItanya: Check that the body gets passed in here as the value?

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -352,6 +398,133 @@ TEST(InjaTransformer, RemoveHeadersUsingEmptyTemplate) {
...
+  dynamic_meta->mutable_value()->set_text("{{body()}}");
+  dynamic_meta = transformation.add_dynamic_metadata_values();
+  dynamic_meta->set_key("bar");
+  dynamic_meta->mutable_value()->set_text("123");
+
+  InjaTransformer transformer(transformation);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> filter_callbacks_{};
+
+  EXPECT_CALL(filter_callbacks_.stream_info_, setDynamicMetadata("io.solo.transformation", _)).Times(2);
```
EItanya: Think it's worth checking that the values are correct as well?

source/extensions/filters/http/transformation/inja_transformer.cc
yuval-k: we actually can! i didnt have the callbacks before.
i will see if i can do the check in the ctor, good idea.

source/extensions/filters/http/transformation/inja_transformer.cc
yuval-k: this only happens one time, i.e. it's not in the request path

source/extensions/filters/http/transformation/inja_transformer.cc
yuval-k: right, but we'll potentially need access to the body, hence why it needs to be here

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: oh duh

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -66,7 +66,8 @@ absl::string_view Extractor::extractValue(absl::string_view value) const {
   if (std::regex_match(value.begin(), value.end(), regex_result,
                        extract_regex_)) {
     if (group_ >= regex_result.size()) {
-      // TODO: this indicates user misconfiguration... log or stat?
+      ASSERT("no such group in the regex");
+      //ENVOY_STREAM_LOG(debug, "invalid group specified for regex", callbacks_);
```
EItanya: Why is this commented out? you can inherit from `Logger::Instance` in these classes to get access. In general I feel that these filters would benefit from more logging

yuval-k: fixed

PR #39, opened 2019-11-05
added unit test for body with newlines

yuval-k:


yuval-k:
/skip-changelog
nothing here changed - just a test added

PR #42, opened 2019-12-03
Make environ explicit in the inja_transformer filter

christian-posta:
BOT NOTES: 
resolves https://github.com/solo-io/envoy-gloo/issues/43


yuval-k:
this fails to due something that will be fixed in the update envoy PR. force merging.
/skip-ci

PR #51, opened 2020-03-03
V1.2.x break

EItanya:


source/common/matcher/matcher.cc
```diff
@@ -54,7 +55,7 @@ class BaseMatcherImpl : public Matcher, public Logger::Loggable<Logger::Id::filt
  */
 class PrefixMatcherImpl : public BaseMatcherImpl {
 public:
-  PrefixMatcherImpl(const ::envoy::api::v2::route::RouteMatch& match)
+  PrefixMatcherImpl(const ::RouteMatch& match)
```
yuval-k: not sure if this will build with the :: prefix in `::RouteMatch`

PR #52, opened 2020-03-10
Clustermeta transformation

kdorosh:
add new command to transformation filter to grab values in the cluster metadata.

Sample cluster metadata:
```
"metadata": {
  "filter_metadata": {
    "io.solo.transformation": {
      "host": "foo.example.com"
     }
  }
}
```

transformation filter usage:
`"{{clusterMetadata("host")}}"`
would result in:
`"foo.example.com"`

### DIFF
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -101,7 +101,7 @@ TransformerInstance::TransformerInstance(
     const Http::HeaderMap &header_map, GetBodyFunc& body,
     const std::unordered_map<std::string, absl::string_view>& extractions,
     const json &context, const std::unordered_map<std::string, std::string>& environ,
-    const envoy::api::v2::core::Metadata* cluster_metadata)
+    const envoy::config::core::v3::Metadata* cluster_metadata)
     : header_map_(header_map), body_(body), extractions_(extractions),
       context_(context), environ_(environ), cluster_metadata_(cluster_metadata) {
   env_.add_callback("header", 1,
@@ -149,7 +149,7 @@ json TransformerInstance::cluster_metadata_callback(const inja::Arguments& args)
     return "";
   }
 
-  const ProtobufWkt::Value& value = Envoy::Config::Metadata::metadataValue(*cluster_metadata_, "io.solo.transformation", key);
+  const ProtobufWkt::Value& value = Envoy::Config::Metadata::metadataValue(cluster_metadata_, SoloHttpFilterNames::get().Transformation, key);
 
   switch (value.kind_case()) {
   case ProtobufWkt::Value::kStringValue: {
@@ -360,7 +360,7 @@ void InjaTransformer::transform(Http::HeaderMap &header_map,
   }
 
   // get cluster metadata
-  const envoy::api::v2::core::Metadata* cluster_metadata{};
+  const envoy::config::core::v3::Metadata* cluster_metadata{};
   Upstream::ClusterInfoConstSharedPtr ci = callbacks.clusterInfo();
   if (ci.get()) {
     cluster_metadata = &ci.get()->metadata();
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -7,6 +7,7 @@
 #include "common/common/macros.h"
 #include "common/common/utility.h"
 #include "common/common/regex.h"
+#include "common/config/metadata.h"
 
 extern char **environ;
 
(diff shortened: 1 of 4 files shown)


source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -132,6 +142,78 @@ json TransformerInstance::env(const inja::Arguments& args) const {
...
 }
 
+json TransformerInstance::cluster_metadata_callback(const inja::Arguments& args) const {
+  const std::string& key = args.at(0)->get_ref<const std::string&>();
+  
+  if (!cluster_metadata_) {
+    return "";
+  }
+
+  const ProtobufWkt::Value& value = Envoy::Config::Metadata::metadataValue(cluster_metadata_, SoloHttpFilterNames::get().Transformation, key);
```
yuval-k: i think this line wont compile. if that's the case accept the suggestion
```suggestion
  const ProtobufWkt::Value& value = Envoy::Config::Metadata::metadataValue(*cluster_metadata_, SoloHttpFilterNames::get().Transformation, key);
```

kdorosh: They very recently changed it:
- the old version took a ref: https://github.com/envoyproxy/envoy/blob/v1.13.1/source/common/config/metadata.h#L39
- the current master (which we are now tracking) takes a pointer: https://github.com/envoyproxy/envoy/blob/master/source/common/config/metadata.h#L45

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -276,8 +358,16 @@ void InjaTransformer::transform(Http::HeaderMap &header_map,
           named_extractor.second.extract(callbacks, header_map, get_body);
     }
   }
+
+  // get cluster metadata
+  const envoy::config::core::v3::Metadata* cluster_metadata{};
+  Upstream::ClusterInfoConstSharedPtr ci = callbacks.clusterInfo();
+  if (ci.get()) {
+    cluster_metadata = &ci.get()->metadata();
```
yuval-k:
```suggestion
    cluster_metadata = &ci->metadata();
```

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -565,6 +601,30 @@ TEST(InjaTransformer, ParseBodyListUsingContext) {
...
+  Http::TestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+  TransformationTemplate transformation;
+  transformation.mutable_body()->set_text("{{clusterMetadata(\"key\")}}");
+
+  InjaTransformer transformer(transformation);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  
+  Upstream::MockClusterInfo* cluster_info = new NiceMock<Upstream::MockClusterInfo>();
+  Upstream::ClusterInfoConstSharedPtr cluster_info_ptr{cluster_info};
```
yuval-k:
```suggestion
Upstream::ClusterInfoConstSharedPtr cluster_info_ptr(new NiceMock<Upstream::MockClusterInfo>());
```

yuval-k: actually you can probably delete this block. see below

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -565,6 +601,30 @@ TEST(InjaTransformer, ParseBodyListUsingContext) {
...
+
+  envoy::config::core::v3::Metadata meta;
+  meta.mutable_filter_metadata()->insert({SoloHttpFilterNames::get().Transformation, MessageUtil::keyValueStruct("key", "val")});
+  ON_CALL(*cluster_info, metadata()).WillByDefault(testing::ReturnRefOfCopy(meta));
```
yuval-k:
```suggestion
  ON_CALL(*callbacks.cluster_info_, metadata()).WillByDefault(testing::ReturnRefOfCopy(meta));
```
and then you can get rid of the cluster_info_ptr and the EXPECT_CALL below

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -565,6 +601,41 @@ TEST(InjaTransformer, ParseBodyListUsingContext) {
...
+  Http::TestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+  TransformationTemplate transformation;
+  transformation.mutable_body()->set_text("{{clusterMetadata(\"key\")}}");
+
+  InjaTransformer transformer(transformation);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  envoy::config::core::v3::Metadata* meta{};
+
+  ON_CALL(*callbacks.cluster_info_, metadata()).WillByDefault(testing::ReturnRef(*meta));
```
yuval-k: *meta is derefing a nil pointer which is techinically undefined behavior.
can we change to?
```
WillByDefault(testing::Return(nullptr));
```

kdorosh: this won't compile, will update with similar

PR #62, opened 2020-05-27
wasm update

yuval-k:


yuval-k:
/skip-changelog
nothing changed; this is just wasm edition

PR #78, opened 2020-08-07
Aws sts conn pool

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/envoy-gloo/issues/80


source/extensions/filters/http/aws_lambda/sts_connection_pool.cc
```diff
@@ -0,0 +1,221 @@
...
+  std::string X;                                                               \
+  {                                                                            \
+    std::match_results<absl::string_view::const_iterator> matched;             \
+    bool result = std::regex_search(body.begin(), body.end(), matched,         \
+                                    StsResponseRegex::get().regex_##X);        \
+    if (!result || !(matched.size() != 1)) {                                   \
+      ENVOY_LOG(trace, "response body did not contain " #X);                   \
+      for (auto &&ctx : connection_list_) {                                    \
+        ctx->callbacks()->onFailure(CredentialsFailureStatus::InvalidSts);     \
+      }                                                                        \
```
yuval-k, line 136: extract this to a method, as this logic is used in dtor too?

we should also empty `(while loop/back()/pop_back())` this list as this may be called again in dtor and onFailure, making the callback invoked twice, similar to how you do it in the bottom of the function.

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -125,6 +125,8 @@ void AWSLambdaFilter::onSuccess(
         credentials) {
   credentials_ = credentials;
   // Do not null context here; all hell will break loose.
```
yuval-k, line 127: i belive this comment is not needed now

source/extensions/filters/http/aws_lambda/config.cc
```diff
@@ -85,15 +102,44 @@ AWSLambdaConfigImpl::AWSLambdaConfigImpl(
...
 }
 
+void AWSLambdaConfigImpl::init() {
+  // Add file watcher for token file
+  auto shared_this = shared_from_this();
+  file_watcher_->addWatch(
+      token_file_, Filesystem::Watcher::Events::Modified,
+      [shared_this](uint32_t) {
+        try {
+          const auto web_token = shared_this->api_.fileSystem().fileReadToEnd(
```
yuval-k, line 112: is this the only place we read the token?
i think we should also do it in the config ctor, so that if the file does not exist, the config should get nack'ed.
wdyt?

source/extensions/filters/http/aws_lambda/config.cc
```diff
@@ -85,15 +102,44 @@ AWSLambdaConfigImpl::AWSLambdaConfigImpl(
   }
 }
 
+void AWSLambdaConfigImpl::init() {
```
yuval-k, line 105: who calls init()? i cant seem to find it?

source/extensions/filters/http/aws_lambda/sts_connection_pool.cc
```diff
@@ -0,0 +1,217 @@
...
+              "Unable to determine expiration time from STS credentials "
+              "result (error: {}), using default",
+              error);
+  }
+
+  StsCredentialsConstSharedPtr result = std::make_shared<const StsCredentials>(
+      access_key, secret_key, session_token, expiration_time);
+
+  // Send result back to Credential Provider to store in cache
+  callbacks_->onSuccess(result, role_arn_);
```
yuval-k, line 164: onSuccess is the only function used in the callback.. maybe change the name onResult?

yuval-k (changes requested):
see commets

source/extensions/filters/http/aws_lambda/config.cc
EItanya: Talked on slack and decided to move the throws into the constructor for the `StsCredentialsProviderFactoryImpl`

yuval-k (approved):

PR #82, opened 2020-09-11
wasm

EItanya:


Sodman:
This was done in https://github.com/solo-io/envoy-gloo/pull/83 instead

PR #96, opened 2020-12-21
[WIP] Allow duplicate header values

saiskee:
Previously, the transformation filter removed multiple headers with the same header name from being processed. This adds a new field `headers_to_append`, which allows users to specify multiple template values for a header, and each of the headers values will be present in the final request.

### DIFF
--- a/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
+++ b/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
@@ -174,6 +174,19 @@ message TransformationTemplate {
   // resulting header, the rendered template will determine the value.
   map<string, InjaTemplate> headers = 3;
 
+  // Defines a header-template pair to be used in `headers_to_append`
+  message HeaderToAppend {
+    // Header name
+    string key = 1;
+    // Apply a template to the header value
+    InjaTemplate value = 2;
+  }
+
+  // Use this attribute to transform request/response headers. It consists of 
+  // an array of string/template objects. Use this attribute to define multiple 
+  // templates for a single header.
+  repeated HeaderToAppend headers_to_append = 10;
+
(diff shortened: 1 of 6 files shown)


api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -174,6 +174,19 @@ message TransformationTemplate {
...
+  message HeaderToAppend {
+    // Header name
+    string key = 1;
+    // Apply a template to the header value
+    InjaTemplate value = 2;
+  }
+
+  // Use this attribute to transform request/response headers. It consists of 
+  // an array of string/template objects. Use this attribute to define multiple 
+  // templates for a single header.
```
EItanya, line 187: Can we explain here that the resulting headers will be added, not replace existing.

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -174,6 +174,19 @@ message TransformationTemplate {
   // resulting header, the rendered template will determine the value.
```
EItanya, line 174: Can we specify that existing headers will be replaced. Can we also explain the behavior if both `headers` and `headers_to_append` is specified.

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -334,6 +334,36 @@ TEST(Transformer, transformSimple) {
   EXPECT_EQ("ABC", headers.get_("x-header"));
 }
 
+TEST(Transformer, transformMultipleHeaderValues) {
```
EItanya, line 337: Can we add a unit test when `headers` and `headers_to_append` is specified.

yuval-k (approved):

PR #104, opened 2021-03-18
Add transformer factory

saiskee:


api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -121,6 +122,8 @@ message Transformation {
     // 'headers', containing the headers, and 'body', containing the original
     // body.
     HeaderBodyTransform header_body_transform = 2;
+    // Configuration for an externally implemented transformer.
```
EItanya, line 125: Any reason not to just have this API be in OSS and mark it as enterprise only. Would avoid the indirection

yuval-k: this is api is not exposed to the user, we can do the OSS approach api in gloo to make it easy on the user.

I don't see an easy way to use this approach here - To do that we will essentially need to override the  transformation config factory from closed source, which i don't think is easy to do with envoy...

test/extensions/filters/http/transformation/transformation_filter_config_test.cc
```diff
@@ -0,0 +1,42 @@
...
+namespace Transformation {
+
+using FakeTransformerProto = envoy::test::extensions::transformation::FakeTransformer;
+
+
+TEST(TransformerExtensionFactory, TestTransformerExtensionFactoryRegistration){
+  envoy::api::v2::filter::http::Transformation transformation;
+  auto factoryConfig = std::make_shared<envoy::config::core::v3::TypedExtensionConfig>();
+  factoryConfig->set_name("io.solo.transformer.fake");
+  transformation.set_allocated_transformer_config(factoryConfig.get());
```
yuval-k, line 35: while i know it's only a test, i would prefer to avoid `set_allocated` funciton familiy in favor for the `mutable_` familiy (more memory safe). i.e.:
```
auto filter_config = transformation.mutable_transformer_config();
filter_config->set_name(...);
```

test/extensions/filters/http/transformation/transformation_filter_config_test.cc
```diff
@@ -0,0 +1,42 @@
...
+  auto &factory = Config::Utility::getAndCheckFactory<TransformerExtensionFactory>(transformation.transformer_config());
+  EXPECT_EQ(factory.name(), "io.solo.transformer.fake");
+}
+}
+}
```
yuval-k, line 40: can we also add a test that covers the `Transformation::getTransformer` code path?

yuval-k (approved):

PR #114, opened 2021-04-26
Access log transformer

jahvon:
This PR adds the `onStreamComplete` callback to the transformation filter. This will allow us to further implement transforming headers that will be logged (to be done in Envoy Gloo EE [here](https://github.com/solo-io/envoy-gloo-ee/pull/148))

### DIFF
--- a/source/extensions/filters/http/transformation/transformation_filter.cc
+++ b/source/extensions/filters/http/transformation/transformation_filter.cc
@@ -27,7 +27,7 @@ TransformationFilter::~TransformationFilter() {}
 
 void TransformationFilter::onDestroy() { resetInternalState(); }
 
-void TransformationFilter::onStreamComplete() { transformAccessLogs(); }
+void TransformationFilter::onStreamComplete() { transformOnStreamCompletion(); }
 
 Http::FilterHeadersStatus
 TransformationFilter::decodeHeaders(Http::RequestHeaderMap &header_map,
@@ -179,8 +179,8 @@ void TransformationFilter::setupTransformationPair() {
         active_transformer_pair->getRequestTranformation();
     response_transformation_ =
         active_transformer_pair->getResponseTranformation();
-    access_log_transformation_ =
-        active_transformer_pair->getAccessLogTransformation();
+    on_stream_completion_transformation_ =
+        active_transformer_pair->getOnStreamCompletionTransformation();
   }
 }
 
@@ -209,8 +209,8 @@ void TransformationFilter::addEncoderData(Buffer::Instance &data) {
   encoder_callbacks_->addEncodedData(data, false);
 }
 
-void TransformationFilter::transformAccessLogs() {
-  if (access_log_transformation_ == nullptr) {
+void TransformationFilter::transformOnStreamCompletion() {
+  if (on_stream_completion_transformation_ == nullptr) {
     return;
   }
 
@@ -219,14 +219,13 @@ void TransformationFilter::transformAccessLogs() {
   // in access logs
   Buffer::OwnedImpl emptyBody{};
   try {
-    access_log_transformation_->transform(*response_headers_,
-                                          request_headers_, 
-                                          emptyBody, 
-                                          *on_complete_callbacks);
-    
+    on_stream_completion_transformation_->transform(*response_headers_,
+                                                    request_headers_, 
+                                                    emptyBody, 
+                                                    *on_complete_callbacks);
   } catch (std::exception &e)  {
     ENVOY_STREAM_LOG(debug, 
-                     "failure transforming access logs {}", 
+                     "failure transforming on stream completion {}", 
                      *on_complete_callbacks, 
                      e.what());
   }
--- a/source/extensions/filters/http/transformation/transformation_filter.cc
+++ b/source/extensions/filters/http/transformation/transformation_filter.cc
@@ -218,11 +218,13 @@ void TransformationFilter::transformOnStreamCompletion() {
   // Body isn't required for this transformer since it isn't included
   // in access logs
   Buffer::OwnedImpl emptyBody{};
+  std::unique_ptr<Http::ResponseHeaderMapImpl> emptyResponseHeaderMap = 
+      Http::ResponseHeaderMapImpl::create();
 
   // If response_headers_ is a nullptr (this can happpen if a client disconnects)
   // we pass in an empty response header to avoid errors within the transformer.
   if (response_headers_ == nullptr) {
-    response_headers_ = Http::ResponseHeaderMapImpl::create().get();
+    response_headers_ = emptyResponseHeaderMap.get();
   }
 
   try {
--- a/source/extensions/filters/http/transformation/transformation_filter.cc
+++ b/source/extensions/filters/http/transformation/transformation_filter.cc
@@ -220,7 +220,7 @@ void TransformationFilter::transformOnStreamCompletion() {
   Buffer::OwnedImpl emptyBody{};
   try {
     on_stream_completion_transformation_->transform(*response_headers_,
-                                                    *request_headers_, 
+                                                    request_headers_, 
                                                     emptyBody, 
                                                     *on_complete_callbacks);
   } catch (std::exception &e)  {
@@ -239,7 +239,7 @@ void TransformationFilter::transformSomething(
     void (TransformationFilter::*addData)(Buffer::Instance &)) {
 
   try {
-    transformation->transform(header_map, *request_headers_, body, callbacks);
+    transformation->transform(header_map, request_headers_, body, callbacks);
 
     if (body.length() > 0) {
       (this->*addData)(body);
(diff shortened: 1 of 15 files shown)


source/extensions/filters/http/transformation/transformation_filter_config.cc
```diff
@@ -202,6 +214,7 @@ void PerStageRouteTransformationFilterConfig::addTransformation(
       TransformerPairConstSharedPtr transformer_pair =
           std::make_unique<TransformerPair>(request_transformation,
                                             response_transformation,
+                                            nullptr,
```
jahvon, line 217: Should access log transformations also be included in StageRouteTransformationFilterConfig? Didn't seem like it when looking at the code but I'm not familiar with eh distinction between this and regular TransformationFIlterConfig.

source/extensions/filters/http/transformation/transformation_filter.cc
```diff
@@ -205,6 +209,29 @@ void TransformationFilter::addEncoderData(Buffer::Instance &data) {
   encoder_callbacks_->addEncodedData(data, false);
 }
 
+void TransformationFilter::transformAccessLogs() {
```
kdorosh, line 212: can this ever be used outside of access logging? name might be a bit too specific

source/extensions/filters/http/transformation/transformation_filter_config.cc
kdorosh: the staged ones are new / supported, the old ones are deprecated. stages allow us to run transformations more than once at different times in the filter chain

EItanya: I'm not sure if DLP supports stages today, the most relevant API for this feature will be that one.

source/extensions/filters/http/transformation/transformation_filter.cc
EItanya: I think I agree, internally we should probably name this onStreamCompleteTransformation, and then we can decide what to call it in the user facing API

source/extensions/filters/http/transformation/transformation_filter_config.cc
kdorosh: Right, I was thinking this was the open source transformation filter when I wrote that comment, not DLP 🤦

source/extensions/filters/http/transformation/transformation_filter.cc
```diff
@@ -205,6 +209,32 @@ void TransformationFilter::addEncoderData(Buffer::Instance &data) {
...
+  Buffer::OwnedImpl emptyBody{};
+  try {
+    on_stream_completion_transformation_->transform(*response_headers_,
+                                                    request_headers_, 
+                                                    emptyBody, 
+                                                    *on_complete_callbacks);
+    on_stream_completion_transformation_->transform(*request_headers_,
+                                                    request_headers_, 
+                                                    emptyBody, 
+                                                    *on_complete_callbacks);
```
EItanya, line 229: I think I see why you're calling twice, but I think using the existing `transform` function is proving to be more complicated than it's worth. Can we instead create a new function to just pass in the necessary data 1 time. 
```
    on_stream_completion_transformation_->transformOnStreamComplete(*request_headers_,
                                                    *response_headers_, 
                                                    *on_complete_callbacks);
```
Or something like that

wdyt @yuval-k

jahvon: Yea, I was thinking the same. It should simplify this if I just had a function that took in everything I needed.

source/extensions/filters/http/transformation/transformer.h
```diff
@@ -48,6 +48,9 @@ class Transformer {
                          const Http::RequestHeaderMap *request_headers,
                          Buffer::Instance &body,
                          Http::StreamFilterCallbacks &callbacks) const PURE;
+  virtual void transform_headers(Http::RequestHeaderMap &,
```
yuval-k, line 51: instead of 2 methods, can we change transform to be
```  virtual void transform(Http::RequestOrResponseHeaderMap &map,
                         // request header map. this has the request header map
                         // even when transforming responses.
                         const Http::RequestHeaderMap *request_headers,
                         absl::optional<std::reference_wrapper<Buffer::Instance>> body,
                         Http::StreamFilterCallbacks &callbacks) 
```
(i.e. change the body to absl::optional<std::reference_wrapper<Buffer::Instance>>)?
this way the impl of the existing transformers doesn't have to change?

jahvon: The body param wasn't the reason behind the change here. I created this new method because I need to pass in both the request and response maps that can be modified by the transform. Prior to having this second method, I had to make two calls to the `transform` method (once to transform the request headers and then once to transform the response header). It seemed a little simpler to pass in everything I needed at once.

I'm not completely familiar with why we are also passing in a constant `request_header` in `transform` but another change I could make would be to remove the const and pass in the response headers as `map` and then request headers into that param so that I have access to modifying them both with one call. Does that sound better @yuval-k?

yuval-k: when this method is called on request, the `request_headers` is nil.
when it is called on the response it is not.
I think we can un-const `request_headers` it. the reason it was passed as const as in the original implementation you never had a reason to change it..

source/extensions/filters/http/transformation/transformation_filter.cc
```diff
@@ -205,6 +209,29 @@ void TransformationFilter::addEncoderData(Buffer::Instance &data) {
...
+  if (on_stream_completion_transformation_ == nullptr) {
+    return;
+  }
+
+  Http::StreamFilterCallbacks *on_complete_callbacks{};
+  // Body isn't required for this transformer since it isn't included
+  // in access logs
+  Buffer::OwnedImpl emptyBody{};
+  try {
+    on_stream_completion_transformation_->transform(*response_headers_,
```
yuval-k, line 222: There might be an edge case where client disconnects, and then response headers will be nullptr (because the stream was completed without a response). we should probably allow response_headers to be nullptr in this case

jahvon: I don't think I follow exactly. Should that be supported at this stage or would we expect it to be handled in the transformer? I think as is, this call will accept a response header that is a nullptr. Are you thinking we shouldn't transform if the client disconnects?

jahvon: Actually I think my assumption was wrong. I'll update the interface to accept nullptr here!

source/extensions/filters/http/transformation/transformation_filter.cc
```diff
@@ -205,6 +209,29 @@ void TransformationFilter::addEncoderData(Buffer::Instance &data) {
   encoder_callbacks_->addEncodedData(data, false);
 }
 
+void TransformationFilter::transformOnStreamCompletion() {
+  if (on_stream_completion_transformation_ == nullptr) {
+    return;
+  }
+
+  Http::StreamFilterCallbacks *on_complete_callbacks{};
```
EItanya, line 217: Do we not have access to the actual stream callbacks? we may need those to mask dynamic metadata?

source/extensions/filters/http/transformation/transformation_filter_config.cc
```diff
@@ -70,10 +71,21 @@ TransformationFilterConfig::TransformationFilterConfig(
               fmt::format("Failed to parse response template: {}", e.what()));
         }
       }
+      if (route_transformation.has_on_stream_completion_transformation()) {
+        try {
+          on_stream_completion_transformation = Transformation::getTransformer(
+              route_transformation.on_stream_completion_transformation(), context);
+        } catch (const std::exception &e) {
+          throw EnvoyException(
+              fmt::format("Failed to parse on stream completion template: {}", e.what()));
```
EItanya, line 80: This error message isn't necessarily accurate. This only applies to the inja transformer, we should make this more generic

source/extensions/filters/http/transformation/transformation_filter.cc
jahvon: There is an encoder and decoder callback. Since we were pass the encode/decode stage I just created a new callback. Do you think those callbacks should be passed in instead?

source/extensions/filters/http/transformation/transformation_filter.cc
jahvon: Instead of updating the interface for `transform` to accept a nullptr, I added in a check that sets `response_headers_` to an empty header map if it's a nullptr. Is that a reasonable solution?

source/extensions/filters/http/transformation/transformation_filter.cc
```diff
@@ -205,6 +210,35 @@ void TransformationFilter::addEncoderData(Buffer::Instance &data) {
...
+  }
+
+  // Body isn't required for this transformer since it isn't included
+  // in access logs
+  Buffer::OwnedImpl emptyBody{};
+
+  // If response_headers_ is a nullptr (this can happpen if a client disconnects)
+  // we pass in an empty response header to avoid errors within the transformer.
+  if (response_headers_ == nullptr) {
+    response_headers_ = Http::ResponseHeaderMapImpl::create().get();
```
yuval-k, line 225: this line is actually undefined behavior; you need to keep the unique_ptr for the duration the variable lives

yuval-k (approved):

PR #124, opened 2021-07-14
Fix extractor regex

saiskee:
If we define a request body as 
```
"here is some body with new lines
id: 1234
"
```
and regex in extractor 
```yaml
myExtractor:
  regex: 'id: ([0-9]{4})'
  subgroup: 1
```
we'd expect the extracted string to be `1234`. However our extractor regex logic doesn't currently work like this. The regex needs to match the **entire** body / input text in order for a match to be found. This is because we use the `std:regex_match` method. I believe `std:regex_search` is closer to the expected behavior of extractor regex.

### DIFF
--- a/test/extensions/filters/http/transformation/inja_transformer_test.cc
+++ b/test/extensions/filters/http/transformation/inja_transformer_test.cc
@@ -227,6 +227,23 @@ TEST(Extraction, ExtractIdFromHeader) {
   EXPECT_EQ("123", res);
 }
 
+TEST(Extraction, ExtractIdFromBody) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"},
+                                         {":authority", "www.solo.io"}};
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  extractor.set_regex("id: ([0-9]{4})");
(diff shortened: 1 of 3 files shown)


test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -227,6 +227,23 @@ TEST(Extraction, ExtractIdFromHeader) {
...
+TEST(Extraction, ExtractIdFromBody) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"},
+                                         {":authority", "www.solo.io"}};
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  extractor.set_regex("id: ([0-9]{4})");
+  extractor.set_subgroup(1);
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body(R"EOF(
+    id: 1234
```
saiskee, line 239: `regex_match` would not work in this case, and this is a very realistic case where we don't want to make the regex match the whole body, but just a substring of it.

saiskee:
Closing out this issue; extractor regex should match the *entire* source input, so this is the behavior we will convey to end-users.

PR #125, opened 2021-07-20
upgrade envoy:

yuval-k:
- fix include path in build files


jahvon (approved):

yuval-k:
parseStdRegexAsCompiledMatcher was removed from upstream, copy pasted it here. see:
https://github.com/envoyproxy/envoy/pull/17061/files

yuval-k:
CompiledStdMatcher was removed here:
https://github.com/envoyproxy/envoy/pull/17193

changelog/v1.19.0-patch1/bump-envoy.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
```
jahvon, line 1: Noticing that the changelog bot is failing. Should this be released as `1.19.0` instead of `1.19.0-patch1`?

source/extensions/filters/http/aws_lambda/BUILD
```diff
@@ -76,8 +76,7 @@ envoy_cc_library(
     repository = "@envoy",
     deps = [
         ":aws_lambda_filter_lib",
-        "@envoy//include/envoy/registry",
```
jahvon, line 79: Am I correct in understanding that we don't need the `registry` dep anymore which is why it's removed here?

yuval-k: we might need it, i removed it because i thought the path changed; it turned out that it didnt. i think it still works because it is included transitivly

changelog/v1.19.0-patch1/bump-envoy.yaml
yuval-k: i think our release name should be patch1 if envoy already released. to indicate that this is based on a released envoy 1.19.0. the reason we don't just use 1.19.0 is that we may have multiple releases based on 1.19.0

@EItanya - i thought we documented this but i cant find where

jahvon (approved):

PR #133, opened 2021-12-06
Upgrade to envoy 1.20

jahvon:


ci/do_ci.sh
```diff
@@ -17,8 +17,8 @@ cp -a $UPSTREAM_ENVOY_SRCDIR/ci/flaky_test $SOURCE_DIR/ci
 
 cp -f $UPSTREAM_ENVOY_SRCDIR/tools/shell_utils.sh $SOURCE_DIR/tools
 
-
-
+# Copy the upstream contrib extensions
+# cp -rf $UPSTREAM_ENVOY_SRCDIR/contrib/ $SOURCE_DIR/contrib/
```
jahvon, line 21:
```suggestion
```

source/common/matcher/solo_matcher.cc
```diff
@@ -120,16 +119,9 @@ class PathMatcherImpl : public BaseMatcherImpl {
 class RegexMatcherImpl : public BaseMatcherImpl {
 public:
   RegexMatcherImpl(const RouteMatch &match) : BaseMatcherImpl(match) {
-    if (match.path_specifier_case() ==
```
jahvon, line 123: This is the main change that I'm unsure is safe. Envoy removed all of these deprecated fields, do we need to continue to support this case?

saiskee: This looks reasonable to me.. we only use safeRegex in the CP for gloo edge. https://github.com/solo-io/solo-projects/blob/b6f702f52a180818b83f966aebb81a30d34d608b/projects/gloo/pkg/plugins/advanced_http/plugin.go#L238

saiskee (approved):

PR #134, opened 2021-12-07
Respect lambda fxn errors and treat as 504

nfuden:
On 200 response with fxn error such as with timeouts treat it as a 504 and pass the 504 along to the caller from downstream.

### DIFF
--- a/changelog/v1.19.0-patch4/bump-envoy.yaml
+++ b/changelog/v1.19.0-patch4/bump-envoy.yaml
@@ -1,5 +1,5 @@
 changelog:
-- type: NEW_FEATURE
+- type: FIX
   issueLink: https://github.com/solo-io/gloo/issues/5618
   resolvesIssue: true
   description: >
--- a/changelog/v1.19.0-patch4/bump-envoy.yaml
+++ b/changelog/v1.19.0-patch4/bump-envoy.yaml
@@ -1,0 +1,6 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/gloo/issues/5618
+  resolvesIssue: true
+  description: >
+    Updates lambda to correctly handle upstream errors that are not encoded in status but rather in the error header.
--- a/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
@@ -126,7 +126,7 @@ Http::FilterHeadersStatus
 AWSLambdaFilter::encodeHeaders(Http::ResponseHeaderMap &headers, bool ) {
 
   if (!headers.get(AWSLambdaHeaderNames::get().FunctionError).empty()){
-    // We treat upstream errors as if it was any other upstream error
+    // We treat upstream function errors as if it was any other upstream error
     headers.setStatus(504);
     return Http::FilterHeadersStatus::Continue;
   }
--- a/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
@@ -164,7 +164,6 @@ void AWSLambdaFilter::onSuccess(
                                        RcDetails::get().CredentialsNotFound);
     return;
   }
- 
   aws_authenticator_.init(access_key, secret_key, session_token);
 
   request_headers_->setReferenceMethod(Http::Headers::get().MethodValues.Post);
@@ -236,7 +235,6 @@ AWSLambdaFilter::decodeTrailers(Http::RequestTrailerMap &) {
 }
 
 void AWSLambdaFilter::lambdafy() {
-  
   handleDefaultBody();
 
   const std::string &invocation_type =
--- a/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
@@ -128,7 +128,6 @@ AWSLambdaFilter::encodeHeaders(Http::ResponseHeaderMap &headers, bool ) {
   if (!headers.get(AWSLambdaHeaderNames::get().FunctionError).empty()){
     // We treat upstream function errors as if it was any other upstream error
     headers.setStatus(504);
-    return Http::FilterHeadersStatus::Continue;
   }
   return Http::FilterHeadersStatus::Continue;
 }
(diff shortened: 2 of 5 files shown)


changelog/v1.19.0-patch4/bump-envoy.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+- type: NEW_FEATURE
```
kdorosh, line 2:
```suggestion
- type: FIX
```

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -120,6 +121,28 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
...
+
+Http::FilterDataStatus AWSLambdaFilter::encodeData(Buffer::Instance &,
+                                                        bool ) {
+  return Http::FilterDataStatus::Continue;
+}
+
+Http::FilterTrailersStatus
+AWSLambdaFilter::encodeTrailers(Http::ResponseTrailerMap &) {
+  return Http::FilterTrailersStatus::StopIteration;
+}
```
kdorosh, line 144: why not move these to the header file like we did with `encode100ContinueHeaders`?

nfuden: Yah makes sense. Do we have a general guideline for complexity allowed in header? Something like if it is anything more than a no-op then dont put it in header?

Namely I am even wondering if even the encode headers should go in to the header file.
My gut feeling is no but while we are having the discussion its good know.

kdorosh: We have generally been following upstream envoy for conventions.

I was just calling out that we were being inconsistent with our approach here, but i don't have a super strong opinion either way

kdorosh:

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -120,6 +121,18 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
   return Http::FilterHeadersStatus::StopIteration;
 }
 
+
+Http::FilterHeadersStatus 
+AWSLambdaFilter::encodeHeaders(Http::ResponseHeaderMap &headers, bool ) {
+
+  if (!headers.get(AWSLambdaHeaderNames::get().FunctionError).empty()){
+    // We treat upstream errors as if it was any other upstream error
```
EItanya, line 129:
```suggestion
    // We treat upstream function errors as if it was any other upstream error
```

test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc
```diff
@@ -402,6 +402,22 @@ TEST_F(AWSLambdaFilterTest, NoCredsAvailable) {
...
+  Http::TestResponseHeaderMapImpl response_headers{
+    {"content-type", "test"},
+    {":method", "GET"},
+    {":authority", "www.solo.io"},
+    {":status", "200"},
+    {"x-amz-function-error", "fakerr"},
+    {":path", "/path"}};
+  auto res = filter_->encodeHeaders(response_headers, true);
+  EXPECT_EQ(Http::FilterHeadersStatus::Continue, res);
+  EXPECT_EQ(response_headers.get_(":status"), "504");
```
kdorosh, line 417: nit, we can use `getStatusValue()` here

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -223,7 +236,7 @@ AWSLambdaFilter::decodeTrailers(Http::RequestTrailerMap &) {
 }
 
 void AWSLambdaFilter::lambdafy() {
-
+
```
kdorosh, line 239: super nit on removing these whitespaces

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -120,6 +121,18 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
...
+Http::FilterHeadersStatus 
+AWSLambdaFilter::encodeHeaders(Http::ResponseHeaderMap &headers, bool ) {
+
+  if (!headers.get(AWSLambdaHeaderNames::get().FunctionError).empty()){
+    // We treat upstream errors as if it was any other upstream error
+    headers.setStatus(504);
+    return Http::FilterHeadersStatus::Continue;
+  }
+  return Http::FilterHeadersStatus::StopIteration;
```
EItanya, line 133: Why are we stopping iteration here?

kdorosh: i think we should always return continue here`

nfuden: Ah I thought I changed that good catch.

kdorosh: i think we should always continue here

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -120,6 +121,18 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
...
 }
 
+
+Http::FilterHeadersStatus 
+AWSLambdaFilter::encodeHeaders(Http::ResponseHeaderMap &headers, bool ) {
+
+  if (!headers.get(AWSLambdaHeaderNames::get().FunctionError).empty()){
+    // We treat upstream function errors as if it was any other upstream error
+    headers.setStatus(504);
+    return Http::FilterHeadersStatus::Continue;
```
kdorosh, line 131: super nit, we can remove this line

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
nfuden: Updated to always send continue and have that continue out of the if.

test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc
nfuden: Updated per request. Definitely shows I grabbed an older example extension for cribbing the test from.

kdorosh (approved):

PR #135, opened 2021-12-10
extensions/filters/http/aws_lambda: New paranoid sts implementation

nfuden:
Sts credentials in some environments were going stale.

This meant that they could pick up the initial web token and had all the access rights but did not detect the updates to the token file. In eks this should be the mounted volume which updates as handled by AWS.

Due to the flakeyness of reproduction and the current filewatcher only having a shallow set of change events this approach attempts to meld the idea of file change watching and a timer. If using a non-aws managed implementation for whatever reason the filewatcher will make sure that we dont miss any changes at a rate faster than normal for aws. The timer will allow for pick up of standard jwt. 

The compromise here comes down the case whereby the token is revoked prior to its normal timeout. We could have up to 14 minutes of failures but is safer than rescraping the file on responses of expired token. The reason here is that if we set a rescrape of the file and the file itself is in a bad state (eks fails to update the token) we could potentially get a scrape of the file per request. Namely gloo request -> proxy, proxy -> sts, proxy -> file, proxy -> sts which could also cause timeouts.

Any thoughts appreciated.


nfuden:
Thought for the future is to make the refresh period configurable but we are leaving it at its current value for now.

nfuden:

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
```diff
@@ -87,4 +87,10 @@ message AWSLambdaConfig {
   // `x-envoy-original-method` headers on the request to AWS lambda. 
   // Defaults to false.
   bool propagate_original_routing = 3;
+
+  // Sets cadence for refreshing credentials for Service Account.
+  // Does nothing if Service account is not set.
+  // Does not affect the default filewatch for service account only augments it.
+  // Defaults to not refreshing on time period. Suggested is 14 minutes.
```
kdorosh, line 94: why 14 instead of something more round like 15 minutes?

kdorosh: fwiw, k8s shared informer has similar model (event driven with polling as backup) and the polling is 10 minutes IIRC

changelog/v1.19.0-patch4/aws-sts-paranoia.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
+- type: FIX
+  issueLink: https://github.com/solo-io/gloo/issues/5253
+  resolvesIssue: false
+  description: >
+    Add new sts refresh timer in addition to filewatch.
```
kdorosh, line 6: won't this still mean up to 14 minutes of failed requests in the event we miss an event on the filewatch? seems worth documenting and making timeout configurable, if so

source/extensions/filters/http/aws_lambda/config.cc
```diff
@@ -140,9 +143,29 @@ void AWSLambdaConfigImpl::init() {
...
+        if (shared_this->credential_refresh_delay_.count() > 0){
+            shared_this->timer_->enableTimer(
+                                shared_this->credential_refresh_delay_);
+        }
+     });
+    if (credential_refresh_delay_.count() > 0){
+          ENVOY_LOG(debug, "{}: STS enabled with {} time refresh",
+       __func__, shared_this->credential_refresh_delay_.count());
+      shared_this->timer_->enableTimer(shared_this->credential_refresh_delay_);
+    }else{
```
kdorosh, line 161: super nit, spacing
```suggestion
    } else {
```

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
nfuden: 14 pulled from existing aws methodology as it polls frequently enough to work with minimal 1hour sessions and their 15 minute overlap

nfuden: I updated to 15 as I dont have strong feelings on this.

yuval-k (approved):

nfuden:

PR #136, opened 2021-12-13
extensions/filters/http/aws_lambda:Add option to keep the original request info

nfuden:
Adds lambda option to persist the original request info that otherwise gets lost.
Set as an option to avoid leaking potentially sensitive path info by default.
Implemented here to allow for usage within access logs.

This would be used in conjunction with access logging in a way like the following:
   
```
access_log: 
            - name: envoy.file_access_log
              typed_config: 
                "@type": "type.googleapis.com/envoy.extensions.access_loggers.file.v3.FileAccessLog"             
                path: /dev/stdout 
                format: '[%START_TIME%] "%REQ(X-ENVOY-ORIGINAL-METHOD?:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%"
%RESPONSE_CODE% %RESPONSE_FLAGS% %BYTES_RECEIVED% %BYTES_SENT% %DURATION%
%RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% "%REQ(X-FORWARDED-FOR)%" "%REQ(USER-AGENT)%"
"%REQ(X-REQUEST-ID)%" "%REQ(:AUTHORITY)%" "%UPSTREAM_HOST%"\n'
```

The format seen above is the default access logging format except with the addition of the original-method header taking precedence over the final Method.


kdorosh (commented):
lgtm!

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -165,6 +165,11 @@ void AWSLambdaFilter::onSuccess(
   }
   aws_authenticator_.init(access_key, secret_key, session_token);
 
+  if (filter_config_->propagateOriginalRouting_==true){
```
kdorosh, line 168: super nit,
```suggestion
  if (filter_config_->propagateOriginalRouting_){
```

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
```diff
@@ -83,4 +83,8 @@ message AWSLambdaConfig {
     google.protobuf.Duration timeout = 3;
   }
 
+    // Set original route information as headers on the request.
+    // If set this information is passed along to AWS.
+    // As a result this can be used for pretty access_logging.
```
kdorosh, line 88:
```suggestion
    // Send downstream method and path as `x-envoy-original-path` and ... 
    // headers on the request to AWS lambda. Defaults to false.
```
more clarity

source/extensions/filters/http/aws_lambda/config.h
```diff
@@ -74,6 +74,7 @@ class AWSLambdaConfig {
   getCredentials(SharedAWSLambdaProtocolExtensionConfig ext_cfg,
                  StsConnectionPool::Context::Callbacks *callbacks) const PURE;
   virtual ~AWSLambdaConfig() = default;
+  bool propagateOriginalRouting_;
```
EItanya, line 77: nit: 
```suggestion
  bool propagate_original_routing_;  
```

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -165,6 +165,11 @@ void AWSLambdaFilter::onSuccess(
   }
   aws_authenticator_.init(access_key, secret_key, session_token);
 
+  if (filter_config_->propagateOriginalRouting_){
```
EItanya, line 168:
```suggestion
  if (filter_config_-> propagate_original_routing_){
```

source/extensions/filters/http/aws_lambda/config.cc
```diff
@@ -43,6 +43,8 @@ AWSLambdaConfigImpl::AWSLambdaConfigImpl(
     : stats_(generateStats(stats_prefix, scope)), api_(api),
       file_watcher_(dispatcher.createFilesystemWatcher()), tls_(tls) {
 
+  propagateOriginalRouting_ = protoconfig.propagate_original_routing().value();
```
EItanya, line 46: `propagate_original_routing_` and should be set above aka:
`propagate_original_routing_(protoconfig.propagate_original_routing().value())`

EItanya (changes requested):

source/extensions/filters/http/aws_lambda/config.cc
kdorosh: oh yes should use the initialization list, agreed

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
kdorosh: agree with this style nit, without the space

source/extensions/filters/http/aws_lambda/config.cc
nfuden: Question, I had tried this previously and I took another shot at rearranging things again but I have the following issue.

This is the Impl of AwsConfig rather than the basic awsconfig. If there is a good way for the check in aws_lambda_filter using the shared abstract AwsConfig to be cast to a AwsConfigImpl then we are good otherwise I dont see how I can use the Impl's initialization list. 

Ill see if I cant find some docs for how c++ works with regards to this.

kdorosh: if you don't find anything helpful feel free to DM me, we can screen share and go over it together. should be very quick

nfuden: Alright so after call with kevin I have now moved this as it was originally being set in a incorrect space. Namely we couldn't inline it here as it was on the virtual class rather than the impl. 
After talking it through it seemed that it was better to move it to the filter instantiation as we had access to the proto rather than adding another virtual method on the config as a getter for the value that would be placed on the impl.

I have now pushed this update.

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
```diff
@@ -83,4 +83,8 @@ message AWSLambdaConfig {
     google.protobuf.Duration timeout = 3;
   }
 
+    // Send downstream path and method as `x-envoy-original-path` and `x-envoy-original-method`
+    // headers on the request to AWS lambda. 
+    // Defaults to false.
```
kdorosh, line 88: this looks good. just wanted to call out the sentiment earlier behind my earlier comment to rewrite this documentation here.

in general it's good to write our proto docs with the following guidance/style: "start with what it does, followed by its default value, and then any callouts or notices for devs/users"

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -55,10 +55,10 @@ const HeaderList AWSLambdaFilter::HeadersToSign =
          Http::Headers::get().ContentType});
 
 AWSLambdaFilter::AWSLambdaFilter(Upstream::ClusterManager &cluster_manager,
-                                 Api::Api &api,
+                                 Api::Api &api, bool should_propagate_origin,
```
kdorosh, line 58: super nit, add new parameter on its own line

kdorosh: after looking further, i see some examples in upstream envoy where they do this too. i think this is fine so long as we abide by the column width rules from upstream

changelog/v1.19.0-rc4/lambda-persist-request-details.yaml
```diff
@@ -0,0 +1,8 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/gloo/issues/5560
+  resolvesIssue: false
+  description: >
+    Adds lambda option to persist the original request info that otherwise gets lost.
+    Set as an option to avoid leaking potentially sensitive path info by default.
```
kdorosh, line 7: include rationale for false by default as well for the mesh case, following example of xff header?

source/extensions/filters/http/aws_lambda/aws_lambda_filter_config_factory.cc
```diff
@@ -26,11 +26,12 @@ AWSLambdaFilterConfigFactory::createFilterFactoryFromProtoTyped(
                                             context.clusterManager()),
       context.dispatcher(), context.api(), context.threadLocal(), stats_prefix,
       context.scope(), proto_config);
-
+  auto should_propagate_origin = proto_config.propagate_original_routing().value();
   return
-      [&context, config](Http::FilterChainFactoryCallbacks &callbacks) -> void {
+      [&context, should_propagate_origin, config](Http::FilterChainFactoryCallbacks &callbacks) -> void {
```
kdorosh, line 31: this looks safe to me. just wanted to call out that this is only safe since we captured by value rather than by reference

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
```diff
@@ -83,4 +83,8 @@ message AWSLambdaConfig {
     google.protobuf.Duration timeout = 3;
   }
 
+  // Send downstream path and method as `x-envoy-original-path` and
+  // `x-envoy-original-method` headers on the request to AWS lambda. 
+  // Defaults to false.
+  google.protobuf.BoolValue propagate_original_routing = 3;
```
kdorosh, line 89: nit, this should be indented further left

kdorosh (changes requested):
after looking more closely at the original proto, i see the field was added to `message AWSLambdaConfig`, so I think the original idea of adding a getter to the interface on that class is cleaner

source/extensions/filters/http/aws_lambda/config.h
```diff
@@ -73,6 +73,7 @@ class AWSLambdaConfig {
   virtual StsConnectionPool::Context *
   getCredentials(SharedAWSLambdaProtocolExtensionConfig ext_cfg,
                  StsConnectionPool::Context::Callbacks *callbacks) const PURE;
+  virtual bool ShouldPropagateOrigin() const PURE;
```
kdorosh, line 76: nit
```suggestion
  virtual bool shouldPropagateOrigin() const PURE;
```

source/extensions/filters/http/aws_lambda/config.h
```diff
@@ -97,6 +98,10 @@ class AWSLambdaConfigImpl
       SharedAWSLambdaProtocolExtensionConfig ext_cfg,
       StsConnectionPool::Context::Callbacks *callbacks) const override;
 
+    bool ShouldPropagateOrigin() const override{
+      return propagate_original_routing_;
```
kdorosh, line 102: nit, let's have the method name match the variable name here (I prefer `propagate_original_routing_`)

EItanya: Also typically cpp methods start with lower case

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
nfuden: I think you may have been looking at an older version when reviewing as I unindented ~ the same time as this review. 
This is indented at 2 spaces in just like the oneof section declaration.
Unless I am missing something here?

kdorosh: looks good to me now 👍 

yeah not sure what happened or what i saw there ¯\_(ツ)_/¯

changelog/v1.19.0-rc4/lambda-persist-request-details.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/gloo/issues/5560
+  resolvesIssue: false
+  description: >
+    Adds lambda option to persist original request info that otherwise is lost.
+    Defaults to false to avoid leaking potentially sensitive info.
+    Follows envoy x-forwarded's example and may need to be false when
```
kdorosh, line 8:
```suggestion
    Similar to envoy's x-forwarded-for header and may be desirable to be false when 
```

nfuden:

changelog/v1.19.0-rc4/lambda-persist-request-details.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
```
kdorosh, line 1: the changelog should be in folder `changelog/v1.19.0-patch4/lambda-persist-request-details.yaml`

kdorosh (changes requested):
nit, update changelog folder for release

kdorosh (commented):
lgtm outside of changelog!

kdorosh:

kdorosh (approved):
still needs @EItanya approval to merge

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
EItanya: Do we care about a null value? If not this should be a primitive bool

kdorosh: how confident are we in the default value here?

> Use wrapped scalar types if there is any potential need for a field to have a default value that does not match the proto3 defaults (0/false/""). For example, new features whose default value may change in the future or security mitigations that should be default safe in the future but are temporarily not enabled.

https://github.com/envoyproxy/envoy/blob/main/api/STYLE.md

nfuden: I thought they were recommending it in general but in retrospect it makes sense as to that being the reason one would use a boolvalue rather than a straight bool.
Will change momentarily.

nfuden:

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
nfuden: Are we then good with keeping this as the primitive and have it be falsey by default? Thoughts?

kdorosh: personally i think it should match xff header for default value and thus should also be `BoolVal` https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-use-remote-address

but i wont make a big stink about it if im outnumbered

EItanya (approved):

PR #138, opened 2021-12-23
Lambda albmode

nfuden:
In this pull we add a per route option to have envoy pretend to be an aws alb for parsing responses from invoked lambdas.

If the option is set on a route then the extension will expect content to come back in json format as specified in https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#respond-to-load-balancer

If the response does not correspond then we respond with a server error (perhaps this response code be configurable at an extension level?)

The only notable change is to be more permissive with regards to mutableheaders combined with headers which alb would normally reject.

Created per: https://github.com/solo-io/gloo/issues/5635


source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -129,9 +130,96 @@ AWSLambdaFilter::encodeHeaders(Http::ResponseHeaderMap &headers, bool ) {
...
   }
+  response_headers_ = &headers;
   return Http::FilterHeadersStatus::Continue;
 }
 
+Http::FilterDataStatus AWSLambdaFilter::encodeData(
+                                      Buffer::Instance &data, bool end_stream ){  
+  if (functionOnRoute() == nullptr || !functionOnRoute()->unwrapAsAlb()){
+    // return response as is if not configured for alb mode
+    return Http::FilterDataStatus::Continue;
```
saiskee, line 141: We found a bug where envoy crashes because encode filters are getting destroyed (due to response body buffer overflow), but `encodeData` is still called on the destroyed filters, in which case if we return `Continue`, it causes a use -after-free. The issue is here: https://github.com/solo-io/gloo/issues/4801.

Since we're adding data to the body here and possibly overflowing the body buffer, this same crash could happen. Let's crash-proof this but checking if `onDestroy` has been called on this filter and returning `Continue` or `StopIteration` accordingly. Here's a PR of how we fixed this in other filters: https://github.com/solo-io/envoy-gloo-ee/pull/159

saiskee (changes requested):

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
nfuden: An interesting point however this particular case should not happen.
I can add comments so that this context is not lost.

I posit this due to the following:
1.  This implementation stops and buffers prior to performing any operations on the data (needed currently due to the json parser). 
2. We create a new buffer to input data in and it is guaranteed to be smaller than the original buffer as we pull subsets. The only transformative step is the optional base64 decoding but base64 encoding is larger than the decoded format so once again only a reduction in size.

Perhaps I am incorrect on this but it seems that we shouldnt have to add the ondestroy fnctionality. 

How would you feel if I added a line next to the buffer and stop section to describe why we are safe.

nfuden: Per slack discussions added the ondestroy to make it easier for future implementors as the overhead is minimal.
https://solo-io-corp.slack.com/archives/C011H3TU17T/p1640283845021400?thread_ts=1640278403.016600&cid=C011H3TU17T

nfuden:
Per discussion with Yuval this will be updated to account for the case where encode data never gets end_stream.

nfuden:
Updated to allow for http2 trailer support and support (per yuval) for buffer overflow / free after use issue (per sai)

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -123,15 +124,137 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
...
+    ENVOY_LOG(debug, "{}: attempted operations while destroyed", __func__);
+    return Http::FilterDataStatus::StopIterationNoBuffer;
+  }
+
+  if (functionOnRoute() == nullptr || !functionOnRoute()->unwrapAsAlb()){
+    // return response as is if not configured for alb mode
+    return Http::FilterDataStatus::Continue;
+  }
+
+  // support for streaming and http2 for future proofing
```
yuval-k, line 156: is this comment still relevant?
i don't think streaming is relevant for lambda, and h2 should already be supported?

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -123,15 +124,137 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
...
+  // Now that the response is finished we know that the following is safe
+  // as the following options will only make the resulting buffer smaller.
+  const Buffer::Instance&  buff = *encoder_callbacks_->encodingBuffer();
+  encoder_callbacks_->modifyEncodingBuffer([this](Buffer::Instance& enc_buf) {
+    Buffer::OwnedImpl body;
+    parseResponseAsALB(*response_headers_, enc_buf, body);
+    enc_buf.drain(enc_buf.length());
+    enc_buf.move(body);
+  });
+  response_headers_->setContentLength(buff.length());
```
yuval-k, line 173: i think this can be extracted to a helper function and re-used in encode trailers?

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -123,15 +124,137 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
...
+void AWSLambdaFilter::parseResponseAsALB(Http::ResponseHeaderMap& headers, 
+                    const Buffer::Instance& json_buf, Buffer::Instance& body) {
+
+  ProtobufWkt::Struct alb_response;                
+  try {
+    MessageUtil::loadFromJson(json_buf.toString(), alb_response);
+  } catch (EnvoyException& ex) {
+    ENVOY_LOG(warn, "{}: alb_unwrap set but did not recieve a json payload", 
+                                                   functionOnRoute()->path());
+    headers.setStatus(static_cast<int>(Http::Code::InternalServerError));
```
yuval-k, line 205: do we want to train the original body here? i see that it is done below?
cover this case with a unit test?

another thing to consider in the futurre: is depending on customer feedback, we may want to add a stat  or dynamic metadata here for production debugging

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -123,15 +124,137 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
...
+    }
+    body.add(rawBody);
+  }
+
+  if (flds.contains("statusCode")){
+    headers.setStatus(flds.at("statusCode").number_value());
+    is_bad_format= is_bad_format || !flds.at("statusCode").has_number_value();
+  }
+  if (flds.contains("headers")){
+    auto hds = flds.at("headers").struct_value();
```
yuval-k, line 228: use `const auto&` to avoid unneeded copying on the datapath (here and below)

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -123,15 +124,137 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
...
+
+
+void AWSLambdaFilter::parseResponseAsALB(Http::ResponseHeaderMap& headers, 
+                    const Buffer::Instance& json_buf, Buffer::Instance& body) {
+
+  ProtobufWkt::Struct alb_response;                
+  try {
+    MessageUtil::loadFromJson(json_buf.toString(), alb_response);
+  } catch (EnvoyException& ex) {
+    ENVOY_LOG(warn, "{}: alb_unwrap set but did not recieve a json payload",
```
yuval-k, line 203: logs here should be debug. Envoy logs on prod should be mostly silent.

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -123,15 +124,137 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
...
+    auto hdsFields = hds.fields();
+    for (auto const& hdrEntry : hdsFields) { 
+      auto list = hdrEntry.second.list_value();
+      for (auto const& val : list.values()){
+        headers.addCopy(Http::LowerCaseString(hdrEntry.first),
+                                                            val.string_value());
+      }
+    }
+  }
+  if (!is_bad_format){
```
yuval-k, line 248: why is the `is_bad_format` done this way? why not just bail early on the first bad format?

yuval-k (commented):
nice work!

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
nfuden: Mainly done so that we can perform proper error handling in a single place.
Changed the flow as the drain can now be handled in the aggregation of the upper level function.

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
nfuden: Ive changed this for all the portions that dont get modified. Which leaves body as we may overwrite it in the case of base64

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
nfuden: Set in finalize()

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
nfuden: lambda currently seems to have stats mainly for credentials. 
If alb mode is failing it would show in route failures. Given that do we want to have a specific alb mode failure stat?
Or does that speak to the metadata approach being better? Perhaps based on how it fails.

Thoughts?

saiskee (approved):
Looks good to me! Thanks Nathan!

PR #142, opened 2022-01-25
feature/lambda sts conn pool

nfuden:
aws lambda sts role chaining (depth 1)

This PR allows for users to chain assume roles so that they can perform cross account lambda invocations. Inside of a single AWS account the prior format worked (as long as trust relationships were correct) to assume any role in any region but did not work for cross account assumption. This PR and most of our sts is targeted for use with EKS but is not exclusive to it.

A chained role is valid for 15m to 1h and is defaulted to 1h which is what we will use for this implementation.
The 1h grace does not care about the expiration for the role used to assume it (see this fun bit for more context https://github.com/hotnops/AWSRoleJuggler/)
The assume role action in sts supports both GET and POST but we use POST as it aligns with our main usage a bit better.

This PR attempts to mash the chained role assumption into our normal fetcher and to correctly link the chained requests to the underlying webtoken assumption request (ie for the role within the host AWS account).


nfuden:
👀  bump 👀

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
```diff
@@ -57,28 +59,59 @@ class StsCredentialsProviderImpl : public StsCredentialsProvider,
...
+    auto conn_pool = connection_pools_.find(chained_role);
+    if (conn_pool != connection_pools_.end()) {
+      conn_pool->second->init(uri_, "", result);
+    }
+    chained_requests.pop_back();
+  }
+}
+
+void StsCredentialsProviderImpl::onFailure(CredentialsFailureStatus status,
+                                      std::list<std::string> chained_requests) {
```
yuval-k, line 103: do we need to modify the original list? if so, we can use reference (&). If not, we can use move (&&).
same for onResult

source/extensions/filters/http/aws_lambda/sts_connection_pool.cc
```diff
@@ -114,64 +128,77 @@ StsConnectionPoolImpl::add(StsConnectionPool::Context::Callbacks *callbacks) {
...
-    if (!result || !(matched.size() != 1)) {                                   \
-      ENVOY_LOG(trace, "response body did not contain " #X);                   \
-      onFailure(CredentialsFailureStatus::InvalidSts);                         \
-      return;                                                                  \
-    }                                                                          \
-    const auto &sub_match = matched[1];                                        \
-    decltype(X) matched_sv(sub_match.first, sub_match.length());               \
-    X = std::move(matched_sv);                                                 \
-  }
+  // using a macro as we need to return on error
```
yuval-k, line 145: its hard to see what changed here with all the white spaces in the diff; can we revert changes that are just whitespace?

nfuden: Sure thing. I have reverted white space changes throughout.

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
nfuden: We dont need to modify the original list but it is the cleanest way. 
I have updated accordingly

yuval-k (approved):

PR #144, opened 2022-02-08
ci: Security posture repication from closed source

nfuden:
Methodology copied from closed source work done by Yuval.

Only novel security posture is the transformation filter as it has an overriding filter.

See closed source https://github.com/solo-io/envoy-gloo-ee/pull/262/files

### DIFF
--- a/security_posture.yaml
+++ b/security_posture.yaml
@@ -1,0 +1,13 @@
+# options are:
+# - data_plane_agnostic
+# - requires_trusted_downstream_and_upstream
+# - robust_to_untrusted_downstream
+# - robust_to_untrusted_downstream_and_upstream
+# - unknown
+extensions:
+- name: filters/http/aws_lambda
+  security_posture: robust_to_untrusted_downstream
+- name: filters/http/nats/streaming
+  security_posture: robust_to_untrusted_downstream
+- name: filters/http/transformation
+  security_posture: robust_to_untrusted_downstream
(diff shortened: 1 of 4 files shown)


security_posture.yaml
```diff
@@ -0,0 +1,13 @@
...
+# - robust_to_untrusted_downstream
+# - robust_to_untrusted_downstream_and_upstream
+# - unknown
+extensions:
+- name: filters/http/aws_lambda
+  security_posture: robust_to_untrusted_downstream
+- name: filters/http/nats/streaming
+  security_posture: robust_to_untrusted_downstream
+- name: filters/http/transformation
+  security_posture: robust_to_untrusted_downstream
```
kdorosh, line 13: is there any guard against forgetting to update this list when a new filter is added?

nfuden: I copied the do_ci and subscript portions from closed source.
In those files we only dont add in the closed source library for filter check 
ie this line is marginally different https://github.com/solo-io/envoy-gloo/blob/39b12d1538886e021c74800c909514ffba137512/ci/verify_posture.sh#L4

As part of this PR i tested that it worked and that the filters are in order as needed. 
In fact in close source we also check this and if a new extension is added here and merged to closed source then we will need to add the filter there as well. 
This is important as some filters may be less hardened in OSS. For example at this point filters/http/transformation is overwritten in closed source and needs its own entry.

kdorosh: I see (both what you said and the ci script updates to ensure devs must update this yaml), makes sense 👍

kdorosh (approved):

PR #149, opened 2022-03-09
fix: aws token not sent when body is added by a filter

yuval-k:
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6023

### DIFF
--- a/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
@@ -153,12 +153,11 @@ Http::FilterDataStatus AWSLambdaFilter::encodeData(
     return Http::FilterDataStatus::Continue;
   }
 
-  encoder_callbacks_->addEncodedData(data, false);
-
   if (!end_stream) {
     // we need the entire response prior to parsing and unwrapping the json
     return Http::FilterDataStatus::StopIterationAndBuffer;
   }
+  encoder_callbacks_->addEncodedData(data, false);
   finalizeResponse();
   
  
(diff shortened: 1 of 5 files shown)


test/integration/aws_lambda_filter_integration_test.cc
```diff
@@ -39,14 +71,71 @@ class AWSLambdaFilterIntegrationTest
...
+      // set env vars for test
+      TestEnvironment::setEnvVar("AWS_WEB_IDENTITY_TOKEN_FILE", TestEnvironment::runfilesPath("test/integration/fakejwt.txt", "envoy_gloo"), 1);
+      TestEnvironment::setEnvVar("AWS_ROLE_ARN", "test", 1);
+      config_helper_.prependFilter(USE_STS_LAMBDA_FILTER);
+      break;
+    case CredMode::DEFAULT:
+      config_helper_.prependFilter(DEFAULT_LAMBDA_FILTER);
+    }
+
+    if (add_transform_) {
```
kdorosh, line 115: why have two separate checks here for `if (add_transform_) {` and then prepend filter? seems like these could live in a single if block

kdorosh (approved):

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -110,7 +110,7 @@ AWSLambdaFilter::decodeHeaders(Http::RequestHeaderMap &headers,
       ENVOY_LOG(trace, "{}: stopping iteration to wait for STS credentials",
                 __func__);
       stopped_ = true;
-      return Http::FilterHeadersStatus::StopAllIterationAndBuffer;
+      return Http::FilterHeadersStatus::StopIteration;
```
kdorosh, line 113: this is the bug fix. we have a filter that adds a body on header-only initial request and thus router filter gets called without appending the required credentials

this is an upstream envoy bug, as with `StopAllIterationAndBuffer` we would expect lambda filter `decodeHeaders` -> `decodeData` -> router filter `encodeHeaders`, but we see that `decodeData` is skipped and thus creds are missing (it races async STS call)

the filter is written in such a way that `StopAllIterationAndBuffer` or `StopIteration` can be used, thus the easier fix is to switch the use here (rather than fix upstream)

kdorosh: fyi @nfuden

PR #169, opened 2022-07-20
initial commit

Bslabe123:


ci/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # This file was inspired by envoy Dockerfile:
 # https://github.com/envoyproxy/envoy/blob/445a67344ffda0c8828c8e438e463fcaa7878434/ci/Dockerfile-envoy-alpine
 
-FROM frolvlad/alpine-glibc@sha256:7b5e8e727246ea48bee1690e30f3fe35925ed9e437f87206870b2ba54fef833f
```
gunnar-solo, line 4: This was originally pinned as a resolution to https://github.com/solo-io/gloo/issues/5980.  Can you check if the new version solves the crashing problem?

gunnar-solo: Slack context, in which I broke down how I'd tested: https://solo-io-corp.slack.com/archives/C02B8KPKYKX/p1646925161140849

PR #174, opened 2022-08-15
Feat/lambda codify unchain

nfuden:
It turns out that it can be quicker / nicer to not chain in some instances such as within an account. 
Allow for aws upstreams to specify this explicitly if they so choose.


nfuden:
/kick https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=cec5ff52-2164-4931-8804-524d8f1d1fc8 somehow the builder is broken?

kdorosh (commented):
lgtm; leaving final review for @EItanya since he has more context

source/extensions/filters/http/aws_lambda/config.cc
```diff
@@ -137,10 +137,12 @@ void AWSLambdaConfigImpl::init(Event::Dispatcher &dispatcher) {
         try {
             const auto web_token = shared_this->api_.fileSystem().fileReadToEnd(
                 shared_this->token_file_);
-            shared_this->tls_.runOnAllThreads(
-                [web_token](OptRef<ThreadLocalCredentials> prev_config) {
-                  prev_config->sts_credentials_->setWebToken(web_token);
-                });
+            if (web_token != "") {
```
EItanya, line 140: If this is empty, should we reject? Is it even worth starting up without this data? Or can it fill in later?

source/extensions/filters/http/aws_lambda/config.h
```diff
@@ -56,6 +56,7 @@ class AWSLambdaProtocolExtensionConfig
     return session_token_;
   }
   const absl::optional<std::string> &roleArn() const { return role_arn_; }
+  const absl::optional<bool> &disableRoleChaining() const { return disable_role_chaining_; }
```
EItanya, line 59: Why is this optional, can't it just be bool? It's either true or false

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
```diff
@@ -152,17 +166,15 @@ StsConnectionPool::Context *StsCredentialsProviderImpl::find(
...
+          .emplace(role_arn_lookup, conn_pool_factory_->build(
               role_arn, this, StsFetcher::create(cm_, api_)))
           .first;
   }
 
-  // Short circuit any additional checks if we are the base arn
-  // Use getInteger as it doesnt check against default runtime and
-  // So that runtime enablement is not decided at per filter runtime
-  // via a proto based flag. 
-  if (role_arn == default_role_arn_  || Runtime::getInteger(
```
EItanya, line 164: Are we dropping support for the runtime val?

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
```diff
@@ -113,21 +114,34 @@ void StsCredentialsProviderImpl::onFailure(CredentialsFailureStatus status,
...
   // If role_arn_arg is present, use that, otherwise use env
   if (role_arn_arg.has_value()) {
     role_arn = role_arn_arg.value();
   }
 
+  std::string role_arn_lookup = role_arn;
+  if (disable_role_chaining) {
+    // if disable_role_chaining is set on an upstream we need a distinction 
+    // so that we can serve both chained and non-chained as needed
+    role_arn_lookup = "no-chain" + role_arn;
```
EItanya, line 136: Super nit: Should we put a separator here?

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
nfuden: Runtime dropped support for us

source/extensions/filters/http/aws_lambda/config.cc
nfuden: Line 123 we check that. This is technically in our timer so this to reject invalid updates (wont fix everything but should help avoid some bad updates). We could also increment a stat if we wanted.

source/extensions/filters/http/aws_lambda/config.h
nfuden: I like the not setting in case we changed it in the future but I think that its actually quite safe to be a bool, I shall change it

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
EItanya: hahahahahaha, were we ever using it?

source/extensions/filters/http/aws_lambda/config.cc
EItanya: Ahhh I see what you mean, yea I think definitely a stat

source/extensions/filters/http/aws_lambda/config.h
```diff
@@ -56,7 +56,7 @@ class AWSLambdaProtocolExtensionConfig
     return session_token_;
   }
   const absl::optional<std::string> &roleArn() const { return role_arn_; }
-  const absl::optional<bool> &disableRoleChaining() const { return disable_role_chaining_; }
+  const i2d_ASN1_BOOLEAN &disableRoleChaining() const { return disable_role_chaining_; }
```
EItanya, line 59: ?

nfuden: Yeah mistyped an autocomplete

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
nfuden: how do you feel about "no-chain-"

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
nfuden: Yeah 1.11s version worked its just in envoy 1.23 that this became locked down

nfuden:
```
uint64_t getInteger(absl::string_view feature, uint64_t default_value) {
  ASSERT(absl::StartsWith(feature, "envoy."));
  if (Runtime::LoaderSingleton::getExisting()) {
    return Runtime::LoaderSingleton::getExisting()->threadsafeSnapshot()->getInteger(
        std::string(feature), default_value);
  }
  ENVOY_LOG_TO_LOGGER(Envoy::Logger::Registry::getLog(Envoy::Logger::Id::runtime), debug,
                      "Unable to use runtime singleton for feature {}", feature);
  return default_value;
}
```

Became 

`uint64_t getInteger(absl::string_view feature, uint64_t default_value) {
  if (absl::StartsWith(feature, "envoy.")) {
    // DO NOT ADD MORE FLAGS HERE. This function deprecated.
    if (feature == "envoy.http.headermap.lazy_map_min_size") {
      return absl::GetFlag(FLAGS_envoy_headermap_lazy_map_min_size);
    }
  }
  if (absl::StartsWith(feature, "re2.")) {
    if (feature == "re2.max_program_size.error_level") {
      return absl::GetFlag(FLAGS_re2_max_program_size_error_level);
    } else if (feature == "re2.max_program_size.warn_level") {
      return absl::GetFlag(FLAGS_re2_max_program_size_warn_level);
    }
  }
  IS_ENVOY_BUG(absl::StrCat("requested an unsupported integer ", feature));
  return default_value;
}`
Which was part of the upstream lockdown on enums and emitting ENVOY_BUG

nfuden:
/kick possible transformation filter flake https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=e35b4f44-6d21-4b0e-bb1a-d56788871af7 

```st/integration/transformation_filter_integration_test.cc:487: Failure
Step #0: Expected equality of these values:
Step #0:   expected_response
Step #0:     Which is: {"headers":{":status":"200","content-length":"0","x-envoy-upstream-service-time":"0"}}
Step #0:   actual_response
Step #0:     Which is: {"headers":{":status":"200","content-length":"0","x-envoy-upstream-service-time":"1"}}
Step #0: Stack trace:
```

nfuden:
/kick because set it multi line reason which our directives dont care for

source/extensions/filters/http/aws_lambda/config.h
```diff
@@ -27,6 +27,8 @@ namespace AwsLambda {
   COUNTER(fetch_failed)                                                        \
   COUNTER(fetch_success)                                                       \
   COUNTER(creds_rotated)                                                       \
+  COUNTER(webtoken_rotated)                                                    \
+  COUNTER(webtoken_failure)                                                    \
```
EItanya, line 31: Do you think this will be clear? I think a gauge here might be more useful. Is this thing ok or not. What do you think?

nfuden: Do you think it should be both? Long running state and current?

EItanya: Good question. I think that could be good. I like more info. So how about:
1. COUNTER: webtoken_rotated
2. COUNTER: empty_webtoken_failure
3. GAUGE: webtoken_state

nfuden: Updated

EItanya (approved):

PR #177, opened 2022-08-23
Wrap as API Gateway 

ben-taussig-solo:
# Description
 - Updates the `AWSLambdaPerRoute` to include a new field, `request_transformer_config`
 - If specified, the lambda filter will attempt to load a transformer from the typeUrl specified at this field
 - The transformer is used to modify the request sent to an upstream Lambda function

# Context
 - See design doc here: https://soloio.slab.com/posts/wrap-as-api-gateway-design-doc-52wpdey8
 - This issue is very similar to the `unwrapAsApiGateway` work completed a couple of weeks ago. 
   - See the issue [here](https://github.com/solo-io/gloo/issues/6143)
   - See the envoy-gloo PR [here](https://github.com/solo-io/envoy-gloo/pull/167)


api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
```diff
@@ -40,6 +40,10 @@ message AWSLambdaPerRoute {
   // used to process response data.
   // This config cannot be configured simultaneously with unwrap_as_alb
```
sam-heilbron, line 41: nit: I imagine that the preferred approach is to no longer use unwrap_as_alb and instead defined a response transformer config to handle that. Can we flag the unwrap_as_alb API as deprecated and include a comment in the proto outlining the preferred way to configure that behavior?

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -265,6 +269,10 @@ bool AWSLambdaFilter::isTransformationNeeded() {
   return functionOnRoute() != nullptr && (functionOnRoute()->unwrapAsAlb() || functionOnRoute()->hasTransformerConfig());
```
sam-heilbron, line 269: nit: Could we update this method to be more explicit: isResponseTransformationNeeded? Now that we have a seprate request and response transformation, I fear it will be easy to misuse this function without understanding its designed for response transformations

sam-heilbron (commented):
This generally looks good! It seems to mirror the existing response transformation, and include support for an identical request transformation. 

What sort of testing can we add to this new field to ensure that we're injecting transformers correctly and handling errors appropriately?

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
ben-taussig-solo: Yeah, definitely makes sense!

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -19,6 +19,10 @@
 
 #include "source/extensions/filters/http/solo_well_known_names.h"
 
+#include "nlohmann/json.hpp"
+// For convenience
+using json = nlohmann::json;
```
sam-heilbron, line 24: How is this used / why was it not needed when we only supported response transformations?

ben-taussig-solo: This is not necessary here -- I was using it when I had a temporary transformer defined here.

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
ben-taussig-solo: Updated that comment here: cb39433a0958a4bec1cae76b686fe1fbbe5ec390
Let me know what you think!

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
ben-taussig-solo: Renamed the `isTransformationNeeded` to `isResponseTransformationNeeded` here: cb39433a0958a4bec1cae76b686fe1fbbe5ec390

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
ben-taussig-solo: Removed this here: cb39433a0958a4bec1cae76b686fe1fbbe5ec390

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
```diff
@@ -29,6 +29,7 @@ message AWSLambdaPerRoute {
   // body is used if the body empty, and an empty body will be sent upstream.
   google.protobuf.StringValue empty_body_override = 4;
 
+  // Deprecated. Use transformer_config to specify an AWS Lambda response transformer instead.
```
sam-heilbron, line 32: nit: It's not really a functional change, but we could also add the deprecated field option: https://developers.google.com/protocol-buffers/docs/proto#options

source/extensions/filters/http/aws_lambda/aws_lambda_filter.h
```diff
@@ -13,6 +13,7 @@
 #include "source/extensions/filters/http/aws_lambda/sts_credentials_provider.h"
 
 #include "api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.pb.validate.h"
+#include "nlohmann/json.hpp"
```
sam-heilbron, line 16: is this necessary?

sam-heilbron (commented):
Small nits. Also, is it possible to add a test around this behavior?

ben-taussig-solo:
Working on a test that creates and configures a dummy transformer!

source/extensions/filters/http/aws_lambda/aws_lambda_filter.h
ben-taussig-solo: Removed this line here: fdf3fa903711b4ad10416c0b7939e1d225295033

api/envoy/config/filter/http/aws_lambda/v2/aws_lambda.proto
ben-taussig-solo: Added the deprecated field option here: fdf3fa903711b4ad10416c0b7939e1d225295033

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -340,7 +345,13 @@ Http::FilterDataStatus AWSLambdaFilter::decodeData(Buffer::Instance &data,
     has_body_ = true;
   }
 
-  aws_authenticator_.updatePayloadHash(data);
+  // If we are not transforming the request, then update the payload hash according to the incoming data
+  // If we are transforming the request, then we will update the payload hash after the transformation
+  if (!isRequestTransformationNeeded()) {
```
nfuden, line 350: question should we just always move the data to request body and then call update with the whole body at finalization?

source/extensions/filters/http/aws_lambda/aws_authenticator.cc
```diff
@@ -60,6 +60,11 @@ void AwsAuthenticator::updatePayloadHash(const Buffer::Instance &data) {
   body_sha_.update(data);
 }
 
+void AwsAuthenticator::resetBodySha() {
```
yuval-k, line 63: this seems to be unused? remove?

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -340,7 +345,13 @@ Http::FilterDataStatus AWSLambdaFilter::decodeData(Buffer::Instance &data,
...
+    aws_authenticator_.updatePayloadHash(data);
+  } else {
+    request_body_.move(data);
```
yuval-k, line 353: we should avoid doing this, as now envoy cannot manage the size of the buffer.

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
yuval-k: see my comment below; we want to let enovy manage the request buffer. we can discuss

yuval-k (changes requested):

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -349,6 +360,9 @@ Http::FilterDataStatus AWSLambdaFilter::decodeData(Buffer::Instance &data,
   }
 
   if (end_stream) {
+    if (isRequestTransformationNeeded() && has_body_) {
```
yuval-k, line 363: why do we check for body here? 
transformations in general do not require a body?

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
ben-taussig-solo: Resolved this here: https://github.com/solo-io/envoy-gloo/pull/177/commits/8b3c855703350391da50ba5677b14c0c7775a6ce

source/extensions/filters/http/aws_lambda/aws_authenticator.cc
ben-taussig-solo: Removed this method here: https://github.com/solo-io/envoy-gloo/pull/177/commits/70e326a0592159c972822643eb51ab8425fbf836

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
ben-taussig-solo: Additionally, updated the tests in response to these changes here: https://github.com/solo-io/envoy-gloo/pull/177/commits/b80e4c155907eab46508d2d49c0176415e7c4039

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
```diff
@@ -392,6 +405,10 @@ void AWSLambdaFilter::lambdafy() {
 void AWSLambdaFilter::handleDefaultBody() {
   if ((!has_body_) && function_on_route_->defaultBody()) {
     Buffer::OwnedImpl data(function_on_route_->defaultBody().value());
+    if (isRequestTransformationNeeded()) {
+      decoder_callbacks_->addDecodedData(data, true);
+      transformRequest();
+    }
```
yuval-k, line 411: if the request is transformed here, would the wrong sha be computed below?
please cover by tests if needed?

source/extensions/filters/http/aws_lambda/config.h
```diff
@@ -180,13 +180,17 @@ class AWSLambdaRouteConfig : public Router::RouteSpecificFilterConfig {
   }
   bool unwrapAsAlb() const { return unwrap_as_alb_; }
   Transformation::TransformerConstSharedPtr transformerConfig() const { return transformer_config_; }
+  Transformation::TransformerConstSharedPtr requestTransformerConfig() const { return request_transformer_config_; }
   bool hasTransformerConfig() const { return has_transformer_config_; }
+  bool hasRequestTransformerConfig() const { return has_request_transformer_config_; }
```
yuval-k, line 185: this can be `request_transformer_config_ != nullptr` and `has_request_transformer_config_` can be removed? that way you don't have 2 variables to keep in sync?

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
yuval-k: please add unit test to prevent regressions

source/extensions/filters/http/aws_lambda/config.h
ben-taussig-solo: Resolved this here: https://github.com/solo-io/envoy-gloo/pull/177/commits/c0e235adc9d34b18061c78df7f7a0d393812e668

source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc
ben-taussig-solo: Clarified this here: https://github.com/solo-io/envoy-gloo/pull/177/commits/5ade735620c3735a894bcb4f5765d4b64e118132
now there is only a single call to `transformRequest` in the lambda filter, in `lambdafy`, after the body has potentially been replaced by the default value

test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
```diff
@@ -0,0 +1,300 @@
...
+  setupRoute(true, false);
+  auto request_headers = setup_encode();
+
+  Buffer::OwnedImpl buf{};
+  auto on_buf_mod = [&buf](std::function<void(Buffer::Instance&)> cb){cb(buf);};
+  EXPECT_CALL(filter_encode_callbacks_, encodingBuffer).WillOnce(Return(&buf));
+  EXPECT_CALL(filter_encode_callbacks_, modifyEncodingBuffer)
+      .WillOnce(Invoke(on_buf_mod));
+
+  auto edResult = filter_->encodeData(buf, true);
```
yuval-k, line 291: I'm surpised you are not hitting this assert
https://github.com/envoyproxy/envoy/blob/7cddecb88ed2588735e1ae3d94a96893b9986b58/source/common/buffer/buffer_impl.cc#L313

to be extra cautious i suggest:
```suggestion
  Buffer::OwnedImpl dataBuf;
  auto edResult = filter_->encodeData(dataBuf, true);
```

test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
```diff
@@ -0,0 +1,300 @@
...
+    transformedAuthorizationHeader[0]->value().getStringView(),
+    headers.get(Http::LowerCaseString("authorization"))[0]->value().getStringView()
+  );
+}
+
+TEST_F(AWSLambdaTransformerTest, TestConfigureResponseTransformer){
+  setupRoute(true, false);
+  auto request_headers = setup_encode();
+
+  Buffer::OwnedImpl buf{};
```
yuval-k, line 285: `{}` is not needed
```suggestion
  Buffer::OwnedImpl buf;
```

test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
```diff
@@ -0,0 +1,300 @@
...
+  // setup to use the request transformer
+  setupRoute(false, true);
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"},
+                                         {":authority", "www.solo.io"},
+                                         {":path", "/getsomething"}};
+  
+  EXPECT_EQ(Http::FilterHeadersStatus::StopIteration,
+            filter_->decodeHeaders(headers, false));
+
+  Buffer::OwnedImpl data("");
```
yuval-k, line 246:
```suggestion
  Buffer::OwnedImpl data;
```

yuval-k (approved):

PR #180, opened 2022-08-30
Feat/lambda chain lifetime

nfuden:


changelog/validation.yaml
```diff
@@ -1,5 +1,5 @@
-relaxSemverValidation: true
```
saiskee, line 1: let's revert this, looks accidental

source/extensions/filters/http/aws_lambda/sts_fetcher.cc
```diff
@@ -97,6 +97,10 @@ class StsFetcherImpl : public StsFetcher,
     // TODO(nfuden) allow for Region this to be overridable. 
     // DefaultRegion is gauranteed to be available but an override may be faster
     aws_authenticator_.sign(&hdrs, HeadersToSign, DefaultRegion);
+    // Log the accessKey but not the secret. This is to show that we have valid 
+    // credentials but does not leak anything secret.
+    ENVOY_LOG(trace, "assume chained [accesskey={}] [sessiontoken={}]",
+                creds->accessKeyId().value(),  creds->sessionToken().value());
```
kdorosh, line 103: session token is also sensitive / should not be logged?

based on:

> If you're making direct HTTPS API requests to AWS, you can sign those requests with the temporary security credentials that you get from the AWS Security Token Service (AWS STS). To do this, you use the access key ID and secret access key that you receive from AWS STS. You use the access key ID and secret access key the same way you would use long-term credentials to sign a request. You also add to your API request the session token that you receive from AWS STS. You add the session token to an HTTP header or to a query string parameter named X-Amz-Security-Token. You add the session token to the HTTP header or the query string parameter, but not both. For more information about signing HTTPS API requests, see [Signing AWS API Requests](https://docs.aws.amazon.com/general/latest/gr/signing_aws_api_requests.html) in the AWS General Reference.

from https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html

kdorosh (commented):
generally lgtm

source/extensions/filters/http/aws_lambda/sts_fetcher.cc
nfuden: Out session token is just the time

nfuden: Never mind you are right I was thinking that this was the sessionRole not the token. Agreed

saiskee (approved):

PR #181, opened 2022-09-02
lambda/authenticator: Verify that we can repeatedly hash

nfuden:


gunnar-solo (approved):
Make it so.

nfuden:
/kick antlv4 tag exists again! but does our cache  know?

PR #182, opened 2022-09-12
Fix/lambda emplace

nfuden:
Update to caching for lambda auth and auth reuse given misordered data when waiting in headers

### DIFF
--- a/source/extensions/filters/http/aws_lambda/aws_authenticator.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_authenticator.cc
@@ -42,6 +42,7 @@ void AwsAuthenticator::init(const std::string *access_key,
   session_token_ = session_token;
   const std::string &secret_key_ref = *secret_key;
   first_key_ = "AWS4" + secret_key_ref;
+  body_sha_ = Sha256{};
 }
 
 AwsAuthenticator::~AwsAuthenticator() {}
--- a/source/extensions/filters/http/aws_lambda/aws_authenticator.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_authenticator.cc
@@ -35,6 +35,9 @@ AwsAuthenticator::AwsAuthenticator(TimeSource &time_source,
   method_ = &Http::Headers::get().MethodValues.Post;
 }
 
+// init sets up some needed tools for the authenticator 
+// but does not clean state and does not need to be the first
+// call to authenticator. Data may be added prior to this call.
 void AwsAuthenticator::init(const std::string *access_key,
                             const std::string *secret_key,
                             const std::string *session_token) {
@@ -42,14 +45,6 @@ void AwsAuthenticator::init(const std::string *access_key,
   session_token_ = session_token;
   const std::string &secret_key_ref = *secret_key;
   first_key_ = "AWS4" + secret_key_ref;
-}
-
-
-void AwsAuthenticator::resetSHA(){
-  // For usage where auth needs to be reused.
-  // This is not in init as data may be added prior to init being called that
-  // should not be blown away.
-  body_sha_ = Sha256{};
 }
 
 AwsAuthenticator::~AwsAuthenticator() {}
--- a/source/extensions/filters/http/aws_lambda/aws_authenticator.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_authenticator.cc
@@ -42,7 +42,6 @@ void AwsAuthenticator::init(const std::string *access_key,
   session_token_ = session_token;
   const std::string &secret_key_ref = *secret_key;
   first_key_ = "AWS4" + secret_key_ref;
-  body_sha_ = Sha256{};
 }
 
 
--- a/source/extensions/filters/http/aws_lambda/aws_authenticator.cc
+++ b/source/extensions/filters/http/aws_lambda/aws_authenticator.cc
@@ -42,6 +42,13 @@ void AwsAuthenticator::init(const std::string *access_key,
   session_token_ = session_token;
   const std::string &secret_key_ref = *secret_key;
   first_key_ = "AWS4" + secret_key_ref;
+}
+
+
+void AwsAuthenticator::resetSHA(){
+  // For usage where auth needs to be reused.
+  // This is not in init as data may be added prior to init being called that
+  // should not be blown away.
   body_sha_ = Sha256{};
 }
 
--- a/source/extensions/filters/http/aws_lambda/aws_lambda_filter.h
+++ b/source/extensions/filters/http/aws_lambda/aws_lambda_filter.h
@@ -81,7 +81,7 @@ public:
   }
   
 
-  const AwsAuthenticator  aws_authenticator() {
+  const AwsAuthenticator  awsAuthenticator() {
     return aws_authenticator_;
   }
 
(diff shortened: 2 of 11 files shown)


source/extensions/filters/http/aws_lambda/aws_lambda_filter.h
```diff
@@ -79,6 +79,11 @@ class AWSLambdaFilter : public Http::StreamFilter,
   const AWSLambdaRouteConfig  * functionOnRoute() {
     return function_on_route_;
   }
+  
+
+  const AwsAuthenticator  aws_authenticator() {
```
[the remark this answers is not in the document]
yuval-k: also, is this used? remove if not

source/extensions/filters/http/aws_lambda/sts_fetcher.cc
```diff
@@ -90,6 +90,8 @@ class StsFetcherImpl : public StsFetcher,
     // Chained assumption specifics 
     const std::string body = fmt::format(StsChainedFormatString, role_arn, now);
     message->body().add(body);
+    // this call resets the sha as nothing should be added prior to this.
```
yuval-k, line 93: if nothing should be added, do we want an ASSERT here?

test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc
```diff
@@ -321,6 +339,64 @@ TEST_F(AWSLambdaFilterTest, SignOnTrailedEndStream) {
...
+  Buffer::OwnedImpl data_2("data");
+  Http::TestRequestTrailerMapImpl trailers_2;
+
+  Buffer::OwnedImpl diff_data("different");
+
+
+
+  // intentionally misorder our on success to simulate a long sts call.
+  filter_->decodeHeaders(headers_2, false);
+  filter_->decodeData(data_2, false);
```
yuval-k, line 390: nice test! is it worth adding one where onsuccess happens after `decodeData(data_2, true);` (with no decodeTrailers) to better simulate the bug we observed? (to clarify, i think it should be in addition to this case as both are valid)

source/extensions/filters/http/aws_lambda/aws_authenticator.cc
```diff
@@ -45,6 +45,14 @@ void AwsAuthenticator::init(const std::string *access_key,
   body_sha_ = Sha256{};
```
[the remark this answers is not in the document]
nfuden: Forgot to commit it from  the merge which reintroduced it hence the failing test in ci

source/extensions/filters/http/aws_lambda/aws_lambda_filter.h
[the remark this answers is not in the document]
nfuden: I only use it in the test to get access to the body sha. Would it be better to friend class it?

source/extensions/filters/http/aws_lambda/sts_fetcher.cc
nfuden: Sure that was more of a paranoia thing on the filter. I will check if its even needed.

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
```diff
@@ -84,12 +84,13 @@ void StsCredentialsProviderImpl::setWebToken(std::string_view web_token) {
...
-    std::list<std::string> &chained_requests) {
-    auto updated = credentials_cache_.emplace(role_arn, result);
-    if (updated.second){
-      credentials_cache_[role_arn] = result;
-    }
+  std::shared_ptr<const StsCredentials> result, std::string role_arn,
+  std::list<std::string> &chained_requests) {
+  auto inserted = credentials_cache_.emplace(role_arn, result);
+  // emplace does not override so check if inserted and if not then overrid
+  if (!inserted.second){
```
yuval-k, line 91: can we add a unit test for this logic?

source/extensions/filters/http/aws_lambda/aws_lambda_filter.h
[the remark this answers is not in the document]
yuval-k: i would add a comment that it is needed for unit tests

source/extensions/filters/http/aws_lambda/sts_fetcher.cc
yuval-k: is the casewhere it is reused and we need to call the function is that in a different place?

test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc
nfuden: It definitely was worth it. Added it and also in the course of it correctly added a context to tell the filter that its running async to sts.

source/extensions/filters/http/aws_lambda/sts_credentials_provider.cc
nfuden: Yep! Added to the full flow test. It calls this function and makes sure that the resulting cached creds are updated.

source/extensions/filters/http/aws_lambda/sts_fetcher.cc
nfuden: This is the case where we are intentionally reusing it to avoid having to recreate it per run. 
That being said I am starting to wonder if we should stop having it as a field on the fetcher at all but rather should just have it live its full lifecycle within this function.

test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc
```diff
@@ -347,54 +359,74 @@ TEST_F(AWSLambdaFilterTest, SignsDataSetByPreviousFilters) {
...
 
   StsConnectionPool::Context::Callbacks *callbackReference;
   EXPECT_CALL(*filter_config_, getCreds).WillRepeatedly(
-     [&](StsConnectionPool::Context::Callbacks *callbacks){
+     [&](StsConnectionPool::Context::Callbacks *callbacks) ->
+                                         StsConnectionPool::Context*{
         callbackReference = callbacks;
+        // context is used for evaluating if we are in async
+        // only other use is cancel so just use stub.
+        return new StsContextStub();
```
yuval-k, line 376: this will leak memory; while not really a problem as this is a test, it will get annoying when we do asan builds to debug future memory issues.
i suggest having it on the stack (near `callbackReference` and returning a pointer to it?)

nfuden: Sure thing. Asan complained about the unused mock but didnt claim that this leaked.

source/extensions/filters/http/aws_lambda/sts_fetcher.cc
```diff
@@ -90,13 +89,17 @@ class StsFetcherImpl : public StsFetcher,
     // Chained assumption specifics 
     const std::string body = fmt::format(StsChainedFormatString, role_arn, now);
     message->body().add(body);
-    aws_authenticator_.init(&creds->accessKeyId().value(), 
+    // this call resets the sha so that our authenticator is in a fresh state
+    // to be reused.
+    auto aws_authenticator = AwsAuthenticator(api_.timeSource(),
+                                             &AWSStsHeaderNames::get().Service);
```
yuval-k, line 95: You can use regular style here:
```suggestion
    AwsAuthenticator aws_authenticator(api_.timeSource(),
                                             &AWSStsHeaderNames::get().Service);
```

yuval-k (approved):

PR #184, opened 2022-09-27
Create Base64 and Substring Transformation Functions

ben-taussig-solo:
# Description
 - Creates `base64_encode` and `base64_decode` Inja functions
 - Creates `substring` Inja function

# Context

 - This functionality was requested in https://github.com/solo-io/gloo/issues/3597
   -  although the issue title just calls out the need for base64 decoding, I have added base64 encoding and substring Inja functions
   - In particular, the use case detailed in the issue comments requires a substring of the decoded base64 string to be extracted, hence the inclusion of the substring function
 - I will be leaving this PR as a draft until I have validated it via Gloo Edge e2e tests
   - Update: These functions are validated here: https://github.com/solo-io/gloo/pull/7255/files
   - These tests pass for me when I run them locally, although it seems like Gloo Edge CI might be having issues pulling the non-released envoy-gloo image

### DIFF
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -132,6 +132,9 @@ TransformerInstance::TransformerInstance(
   env_.add_callback("base64Decode", 1, [this](Arguments &args) {
     return base64_decode_callback(args); 
   });
+  env_.add_callback("substring", 3, [this](Arguments &args) {
+    return substring_callback(args); 
+  });
 }
 
 json TransformerInstance::header_callback(const inja::Arguments &args) const {
@@ -258,6 +261,30 @@ json TransformerInstance::base64_encode_callback(const inja::Arguments &args) co
 json TransformerInstance::base64_decode_callback(const inja::Arguments &args) const {
   const std::string &input = args.at(0)->get_ref<const std::string &>();
   return Base64::decode(input);
+}
+
+json TransformerInstance::substring_callback(const inja::Arguments &args) const {
+  const std::string &input = args.at(0)->get_ref<const std::string &>();
+  const int64_t start = args.at(1)->get_ref<const int64_t &>();
+  const int64_t substring_len = args.at(2)->get_ref<const int64_t &>();
+
+  // store input length 
+  const int64_t input_len = input.length();
+
+  // if start is negative, substring_len is non-positive, or 
+  // start is greater than the length of the string, return empty string
+  if (start < 0 || substring_len <= 0 || start >= input_len) {
+    return "";
+  }
+
+  // if start + len is greater than the length of the string, return the substring
+  // from start to the end of the string
+  if (start + substring_len > input_len) {
+    return input.substr(start);
+  }
+
+  // otherwise, return the substring from start to start + len
+  return input.substr(start, substring_len);
 }
 
 std::string TransformerInstance::render(const inja::Template &input) {
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -274,8 +274,10 @@ json TransformerInstance::base64_decode_callback(const inja::Arguments &args) co
 // and extending for length characters. If length is not provided, the
 // substring will extend to the end of the string.
 json TransformerInstance::substring_callback(const inja::Arguments &args) const {
+  // get first argument, which is the string to be modified
   const std::string &input = args.at(0)->get_ref<const std::string &>();
-  // try to get first argument (start position) as an int64_t
+  
+  // try to get second argument (start position) as an int64_t
   int start = 0;
   try {
     start = args.at(1)->get_ref<const int64_t &>();
@@ -284,7 +286,7 @@ json TransformerInstance::substring_callback(const inja::Arguments &args) const 
     return "";
   }
 
-  // try to get optional substring_len argument
+  // try to get optional third argument (length) as an int64_t
   int64_t substring_len = -1;
   if (args.size() == 3) {
     try {
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -132,6 +132,12 @@ TransformerInstance::TransformerInstance(
   env_.add_callback("base64_decode", 1, [this](Arguments &args) {
     return base64_decode_callback(args); 
   });
+  // Substring can be called with either two or three arguments --
+  // the first argument is the string to be modified, the second is the start position
+  // of the substring, and the optional third argument is the length of the substring.
+  env_.add_callback("substring", 2, [this](Arguments &args) {
+    return substring_callback(args); 
+  });
   env_.add_callback("substring", 3, [this](Arguments &args) {
     return substring_callback(args); 
   });
@@ -266,7 +272,11 @@ json TransformerInstance::base64_decode_callback(const inja::Arguments &args) co
 json TransformerInstance::substring_callback(const inja::Arguments &args) const {
   const std::string &input = args.at(0)->get_ref<const std::string &>();
   const int64_t start = args.at(1)->get_ref<const int64_t &>();
-  const int64_t substring_len = args.at(2)->get_ref<const int64_t &>();
+  // get optional substring_len argument
+  int64_t substring_len = -1;
+  if (args.size() == 3) {
+    substring_len = args.at(2)->get_ref<const int64_t &>();
+  }
   const int64_t input_len = input.length();
 
   // if start is negative, or start is greater than the length of the string, return empty string
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -267,19 +267,16 @@ json TransformerInstance::substring_callback(const inja::Arguments &args) const 
   const std::string &input = args.at(0)->get_ref<const std::string &>();
   const int64_t start = args.at(1)->get_ref<const int64_t &>();
   const int64_t substring_len = args.at(2)->get_ref<const int64_t &>();
-
-  // store input length 
   const int64_t input_len = input.length();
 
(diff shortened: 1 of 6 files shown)


source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -244,6 +253,40 @@ json TransformerInstance::cluster_metadata_callback(
...
+  const std::string &input = args.at(0)->get_ref<const std::string &>();
+  const int64_t start = args.at(1)->get_ref<const int64_t &>();
+  const int64_t substring_len = args.at(2)->get_ref<const int64_t &>();
+
+  // store input length 
+  const int64_t input_len = input.length();
+
+  // if start is negative, substring_len is non-positive, or 
+  // start is greater than the length of the string, return empty string
+  if (start < 0 || substring_len <= 0 || start >= input_len) {
```
kdorosh, line 276: should we handle substring_len differently if it's zero? e.g. do what the c++ function does and just take until the end? otherwise it seems like the UX is a bit poor for users who want to take substring from start to end without specifying the length

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -126,6 +126,15 @@ TransformerInstance::TransformerInstance(
   env_.add_callback("clusterMetadata", 1, [this](Arguments &args) {
     return cluster_metadata_callback(args);
   });
+  env_.add_callback("base64Encode", 1, [this](Arguments &args) {
+    return base64_encode_callback(args); 
+  });
+  env_.add_callback("base64Decode", 1, [this](Arguments &args) {
+    return base64_decode_callback(args); 
+  });
+  env_.add_callback("substring", 3, [this](Arguments &args) {
```
kdorosh, line 135: curious if we can make the third arg here optional, which seems like a better UX to me

ben-taussig-solo: I'll look into it -- I think we might get a render error if we don't get exactly 3 arguments, but It would be nice to have if that doesn't happen

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -244,6 +253,40 @@ json TransformerInstance::cluster_metadata_callback(
...
 }
 
+json TransformerInstance::base64_encode_callback(const inja::Arguments &args) const {
+  const std::string &input = args.at(0)->get_ref<const std::string &>();
+  return Base64::encode(input.c_str(), input.length());
+}
+
+json TransformerInstance::base64_decode_callback(const inja::Arguments &args) const {
+  const std::string &input = args.at(0)->get_ref<const std::string &>();
+  return Base64::decode(input);
```
kdorosh, line 263: do you know if the use case requires / we may need to expose the `remove_linebreaks` option here? https://renenyffenegger.ch/notes/development/Base64/Encoding-and-decoding-base-64-with-cpp/

kdorosh (commented):
looks good overall! just UX / exposing more behavior style comments

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: https://github.com/solo-io/envoy-gloo/pull/184/commits/91225d5a1a4b1c320c19f6b964f45ce23ffe8256
Updated this here, let me know what you think!

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Did this here: https://github.com/solo-io/envoy-gloo/pull/184/commits/ef3a4251bf97736297be9bec0bdff14b9d5a4ed2

We just had to register two `substring` callbacks, one with two arguments and one with three

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: I think it probably does not (seems like they need to decode an auth token from some external service?), but it's trivial to expose an optional second parameter to the user so I'm going to default `remove_linebreaks` to false and give the user the option to enable it

ben-taussig-solo: Ah, so the base64 utilities that we're using here (defined in upstream Envoy) only support a single argument to `decode`, see here: https://github.com/envoyproxy/envoy/blob/main/source/common/common/base64.cc#L146

I think it's pretty reasonable to assume that the string being decoded does not contain linebreaks -- what do you think?

kdorosh: I think it's fine to leave as is until a request for it comes in.. if envoy hasn't needed it yet then i doubt we will

source/extensions/filters/http/transformation/inja_transformer.cc
ashishb-solo: what are some typical ways that negative numbers are used for substringing? i know that in python, slicing with negative values uses indices from the end, for example:

```
>>> test_str = '0123456789'
>>> print(test_str[-5:-2]) # start from 5 before end, end at 2 before end
567
```

but in python's case, the indices mean `start` and `end` rather than `start` and `length`. i don't have experience with substringing using `start` and `length`. but it does feel a bit redundant to do the same thing with all negative values, and i wonder if we could do something more flexible with them. any thoughts?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -244,6 +259,41 @@ json TransformerInstance::cluster_metadata_callback(
...
+}
+
+json TransformerInstance::substring_callback(const inja::Arguments &args) const {
+  const std::string &input = args.at(0)->get_ref<const std::string &>();
+  const int64_t start = args.at(1)->get_ref<const int64_t &>();
```
kdorosh, line 274: can we add a test for what happens here if a user provides a non-integer arg?

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: The implementation exposed in this PR is pretty similar to the C++ `<string>.substr` function that we call into in the `substring_callback`
https://github.com/solo-io/envoy-gloo/blob/8ddf225f48d8db0fc7f86aa9c62411797967568e/source/extensions/filters/http/transformation/inja_transformer.cc#L294
where the arguments to the substring function are 
1. the nonnegative start position of the substring
2. (optional) the number of characters in the substring

(We also sometimes use the Google Abseil version of this which operates on a `string_view` and has the same style of signature, see here:
https://github.com/solo-io/envoy-gloo/blob/8ddf225f48d8db0fc7f86aa9c62411797967568e/source/common/matcher/solo_matcher.cc#L99

I'm in favor of maintaining this implementation instead of switching to the python style negative substringing for a couple of reasons:
1. Because we don't (at least as far as I know) have a substring utility that we can call into with python style arguments, implementing the python style substringing is going to be more verbose and therefore harder to debug/maintain going forward than just calling into the available C++ utility
2. I don't think the use case that we're implementing this feature for requires a particularly sophisticated set of substringing tools -- I don't have any info beyond what is given in the comments on the issue, but I would imagine that the workflow is generally something like:
  - decode the base64 string, resulting in a constant length auth token
  - remove a certain known number of characters from the start or end of the string

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: good call, will do

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Sorry if that's not a super direct answer. Long story short, I prefer keeping the implementation simple/close to the C++ utilities we're using because 
 - I think it will be easier to maintain
 - it doesn't seem like the feature requester requires a particularly sophisticated set of substringing tools

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Added this here: https://github.com/solo-io/envoy-gloo/pull/184/commits/7e36cc2559e7a994f249aecd1f8276d67375cd3c

A couple of notes
 - The pattern in the custom Inja callbacks we define here seems to be that we return an empty string if there is invalid input/we can't find what the input points to. See for example: https://github.com/solo-io/envoy-gloo/blob/8ddf225f48d8db0fc7f86aa9c62411797967568e/source/extensions/filters/http/transformation/inja_transformer.cc#L169-L177
   - I think this makes sense intuitively, because the callbacks are supposed to be composable, so returning an error string or throwing an exception is probably not desirable behavior
 - The substring callback now returns an empty string if the second argument cannot be parsed as an integer
 - Probably more open to interpretation, the substring callback now also returns an empty string if there is a third argument and it cannot be parsed as an integer
   - How do you feel about this? Do you think we should just act as though there is no third argument in this case?

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Sounds good. Resolving this thread for now

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Also, the envoy base64 utilities we call into will return an empty string if we attempt to decode invalid base64. I'm going add a test against this as well
https://github.com/envoyproxy/envoy/blob/main/source/common/common/base64.cc#L146-L151

ben-taussig-solo: Added this here: f105915196798ca381358177f2c9e34070cf2919

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Closing this thread after our discussion in the Gloo Slack:
https://solo-io-corp.slack.com/archives/C03MFATU265/p1664386328617359

kdorosh (approved):

PR #186, opened 2022-10-07
bazel: bump to envoy patch v1.21.5

ashleywang1:


nfuden:
I think the changelog needs to be changed. 
I state this because our semver libraries assume that if here is a label on a semver then it is likely a prerelease and therefore lower than the standard version. Given that we increment ${envoy_version}-patch${our_changecount} I think this could cause issues down the line.

I would vote for either 1.21.5-patch1 or 1.21.5-patch3

nfuden (approved):

PR #187, opened 2022-10-14
Fix: wrapAsApiGateway crashes Envoy on headers-only request

ben-taussig-solo:
# Description

- Resolves a bug which caused envoy to crash when a headers-only request was sent to an AWS Lambda upstream with `wrapAsApiGateway` enabled.
- This occurred because we previously called `decoder_callbacks_->modifyDecodingBuffer` whenever `wrapAsApiGateway` was enabled
  - Because a decoding buffer does not exist on a headers-only request, attempting to modify a nonexistent decoding buffer caused Envoy to crash
- Now, we transform an empty buffer and then create a decoding buffer from the transformed result using `decoder_callbacks_->addDecodedData` in the case of a headers only request

# Context

- The original issue reporting this bug can be found here: https://github.com/solo-io/gloo/issues/7307
- I will put up and link a solo-projects draft PR with e2e tests that validate this fix end-to-end
  - Update: that PR can be found here https://github.com/solo-io/solo-projects/pull/4177
- That PR will eventually include an envoy-gloo-ee version bump which pulls in the changes contained in this PR

### DIFF
--- a/test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
+++ b/test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
@@ -254,7 +254,7 @@ TEST_F(AWSLambdaTransformerTest, TestConfigureRequestTransformerSignatureNoBody)
 
 
   EXPECT_EQ(transformedxAmzDateHeader[0]->value().getStringView(), "20010909T014640Z");
-  EXPECT_EQ(transformedAuthorizationHeader[0]->value().getStringView(), "AWS4-HMAC-SHA256 Credential=access key/20010909/us-east-1/lambda/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-invocation-type;x-amz-log-type, Signature=6ac2a229351a686ff5a93deb46896f2128a6800a885ef5324c0c0685aa21d786");
+  EXPECT_EQ(transformedAuthorizationHeader[0]->value().getStringView(), "AWS4-HMAC-SHA256 Credential=access key/20010909/us-east-1/lambda/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-invocation-type;x-amz-log-type, Signature=561409e1250c56044b11a3d6eaed9f3d3d9467f3214dfec4786a65381bdab23e");
 
   // now, setup to use no transformer
   setupRoute(false, false);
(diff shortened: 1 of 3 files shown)


test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
```diff
@@ -254,7 +254,7 @@ TEST_F(AWSLambdaTransformerTest, TestConfigureRequestTransformerSignatureNoBody)
 
 
   EXPECT_EQ(transformedxAmzDateHeader[0]->value().getStringView(), "20010909T014640Z");
-  EXPECT_EQ(transformedAuthorizationHeader[0]->value().getStringView(), "AWS4-HMAC-SHA256 Credential=access key/20010909/us-east-1/lambda/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-invocation-type;x-amz-log-type, Signature=6ac2a229351a686ff5a93deb46896f2128a6800a885ef5324c0c0685aa21d786");
+  EXPECT_EQ(transformedAuthorizationHeader[0]->value().getStringView(), "AWS4-HMAC-SHA256 Credential=access key/20010909/us-east-1/lambda/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-invocation-type;x-amz-log-type, Signature=561409e1250c56044b11a3d6eaed9f3d3d9467f3214dfec4786a65381bdab23e");
```
ben-taussig-solo, line 257: Note to reviewers: I'm not sure why this hash has changed. I am actively looking into it now. My testing indicates that requests are being accepted by Amazon (I will link solo-projects e2e tests that validate this), so I don't think this is problematic, but I am currently confused by _why_ it is occurring

kdorosh (commented):
lgtm pending the hash change

ben-taussig-solo:
Update:
 - I created a solo-projects PR [here](https://github.com/solo-io/solo-projects/pull/4177) which adds testing to validate these changes
 - I made a slight change to the order of operations in the headers-only request case [here](https://github.com/solo-io/envoy-gloo/pull/187/commits/be9b62b29b37dcd75016bd218fd2bd595f7097da), as we cannot access the buffer after it has been used to create the `decodingBuffer`
   - This was preventing us from setting the `content-length` header appropriately
 - I am still not sure why the hash has changed in that seemingly unrelated test. Requests with and without request bodies are being accepted and processed normally by AWS using the current changeset, so I don't think this is reason for concern, but I think it's valuable to make note of this

test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
ben-taussig-solo: So this happened because this test was actually hitting the bug that this PR resolves (should've known from the name `TestConfigureRequestTransformerSignatureNoBody`)

On master, running this tests confirms that the "transformed" data buffer used in this test is just an empty string. Now that that buffer contains the expected value, the hash of that value has changed.

kdorosh (approved):
nice job figuring out root cause! :rocket:

PR #193, opened 2022-12-01
Updates for Envoy v1.24.0

chunter0:


test/extensions/filters/http/transformation/transformation_filter_test.cc
```diff
@@ -230,7 +230,7 @@ class TransformationFilterTest : public testing::Test {
     int decoded_data_calls = val == 0 ? 0 : 1;
     EXPECT_CALL(filter_callbacks_, addDecodedData(_, false))
         .Times(decoded_data_calls);
-    EXPECT_CALL(filter_callbacks_, clearRouteCache()).Times(0);
+    EXPECT_CALL(filter_callbacks_, downstreamCallbacks()->clearRouteCache()).Times(0);
```
chunter0, line 233: @EItanya would you happen to know how to generate these for mocks?

EItanya: I'm not sure they're generated for these cpp mocks, mostly we use the upstream ones, there may be one available for this

EItanya (commented):
LGTM, just want confirmation from @nfuden or @kdorosh about versioning in changelog

changelog/v1.24.0-patch1/upgrade-envoy.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
```
kdorosh, line 1: should we go to 1.24.1 https://www.envoyproxy.io/docs.html ? rather than 1.24.0

kdorosh: since 1.24 has been released, i would expect this to be a patch release rather than rc (im assuming the upstream envoy commit is 1.24.0 or newer)

kdorosh:
when this merges we will need to create a new branch for the 1.23 release envoy branch to track based on the last commit before this merged

.bazelrc
```diff
@@ -65,6 +66,9 @@ build:clang --linkopt=-fuse-ld=lld
 build:clang-pch --spawn_strategy=local
 build:clang-pch --define=ENVOY_CLANG_PCH=1
 
+# Use gold linker for gcc compiler.
+build:gcc --linkopt=-fuse-ld=gold
```
kdorosh, line 70: :tada: this is nice it should be faster :100:

kdorosh: to link, not run, i mean. but at least dev flow is faster

kdorosh (approved):
istio 1.16.0 is only at 1.24.0. hence not going to 1.24.1

changelog/v1.24.0-patch1/upgrade-envoy.yaml
kdorosh: istio 1.16.0 is only at 1.24.0. hence not going to 1.24.1

PR #196, opened 2023-01-12
AWS Lambda: Fix issues with request metadata when processing request transformation

ben-taussig-solo:
# Description
 - This PR resolves an issue when processing request transformations in the AWS lambda filter.
   - Prior to this change, the AWS lambda filter would modify the `:path` and `:method` pseudo-headers before the request transformation was processed, resulting in the upstream receiving invalid values of these headers.
 - Once this change is pulled into enterprise, users will be receive valid path/querystringparams/method values when configuring `wrapAsApiGateway`

# Implementation Details
 - Adds a new helper, `AWSLambdaFilter::updateHeaders`, which is used to ensure that all headers which need to be modified for the sake of routing to AWS can be modified in the same place. This method is called immediately after request transformations are processed
 - The test transformer used in `aws_lambda_transformer_test.cc` now places the headers it receives in the request body, allowing us to validate the headers used by request (and response) transformations
   - Because of this, tests which expect a specific signature to be outputted from this transformer need to be updated, as the signature is computed directly from the request body sent to AWS

### DIFF
--- a/test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
+++ b/test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
@@ -189,6 +189,10 @@ TEST_F(AWSLambdaTransformerTest, TestConfigureRequestTransformer){
   EXPECT_EQ(Http::FilterDataStatus::Continue, filter_->decodeData(data, true));
   // Confirm that the body contains the expected value as a substring
   EXPECT_THAT(upstream_body, testing::HasSubstr("test body from fake transformer"));
+  // Confirm that the headers are present in the response body
+  EXPECT_THAT(upstream_body, testing::HasSubstr(":method; GET"));
+  EXPECT_THAT(upstream_body, testing::HasSubstr(":authority; www.solo.io"));
+  EXPECT_THAT(upstream_body, testing::HasSubstr(":path; /getsomething"));
 }
 
 TEST_F(AWSLambdaTransformerTest, TestConfigureRequestTransformerSignature){
--- a/test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
+++ b/test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
@@ -187,8 +187,8 @@ TEST_F(AWSLambdaTransformerTest, TestConfigureRequestTransformer){
       }));
 
   EXPECT_EQ(Http::FilterDataStatus::Continue, filter_->decodeData(data, true));
-  // Confirm that the body is transformed to the hardcoded string.
-  EXPECT_EQ(upstream_body, "test body from fake transformer");
+  // Confirm that the body contains the expected value as a substring
+  EXPECT_THAT(upstream_body, testing::HasSubstr("test body from fake transformer"));
 }
 
 TEST_F(AWSLambdaTransformerTest, TestConfigureRequestTransformerSignature){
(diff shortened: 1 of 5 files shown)


test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
```diff
@@ -254,7 +258,7 @@ TEST_F(AWSLambdaTransformerTest, TestConfigureRequestTransformerSignatureNoBody)
 
 
   EXPECT_EQ(transformedxAmzDateHeader[0]->value().getStringView(), "20010909T014640Z");
-  EXPECT_EQ(transformedAuthorizationHeader[0]->value().getStringView(), "AWS4-HMAC-SHA256 Credential=access key/20010909/us-east-1/lambda/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-invocation-type;x-amz-log-type, Signature=561409e1250c56044b11a3d6eaed9f3d3d9467f3214dfec4786a65381bdab23e");
+  EXPECT_EQ(transformedAuthorizationHeader[0]->value().getStringView(), "AWS4-HMAC-SHA256 Credential=access key/20010909/us-east-1/lambda/aws4_request, SignedHeaders=content-type;host;x-amz-date;x-amz-invocation-type;x-amz-log-type, Signature=31e8a35e5818a5a1b969bc5d80e48d10f5478c12d67cbbace4472a1566ede502");
```
ben-taussig-solo, line 261: Because the test transformer has changed, so has the signature computed from its output

ben-taussig-solo:
Leaving the work-in-progress tag on this PR until I have validated these changes in envoy-gloo-ee and solo-projects. 

At the time of writing this comment, I have this PR in envoy-gloo-ee to validate these changes: https://github.com/solo-io/envoy-gloo-ee/pull/471

jbohanon (commented):
This looks good to me!

test/extensions/filters/http/aws_lambda/test_transformer.h
```diff
@@ -13,12 +13,23 @@ class FakeTransformer : public HttpFilters::Transformation::Transformer {
...
                          Http::RequestHeaderMap *,
                          Buffer::Instance &body,
                          Http::StreamFilterCallbacks &) const override {
+                            std::string *headers_string = new std::string("headers:\n");
+                            headers.iterate(
+                                [headers_string](const Http::HeaderEntry &header) -> Http::HeaderMap::Iterate {
+                                    auto key = std::string(header.key().getStringView());
+                                    auto value = std::string(header.value().getStringView());
+                                    // use semicolon as a separator, because pseudo-headers (e.g. :path) have colons (":") in them
+                                    *headers_string += "\t" + key + "; " + value + "\n";
```
ashishb-solo, line 26: Should we be concerned about performance here? I'm not sure how string appends work in C++, but in some languages, each append is involves a memory allocation and copy. Suggestion to consider using a string buffer or a join?

ashishb-solo: This might be a safer solution https://stackoverflow.com/a/16029964

I think the only cause for concern here is if `HeaderMap` either has many strings or large ones. If so, then we might want to consider the approach above. If not, then maybe this is just a premature optimisation and we don't need to worry about it.

ben-taussig-solo: Good callout, but I think this is not super relevant because this transformer is only used during tests -- in this case I think it makes sense to prioritize readability over a handful of milliseconds so that our tests are easier to work with/understand

ashishb-solo: ah, didn't see that this was in tests. fine by me then

yuval-k (approved):
lgtm

PR #198, opened 2023-01-19
upgrade to envoy 1.25

saiskee:
Update to upstream 1.25
https://github.com/envoyproxy/envoy/blob/9184b84cd0dcb3a6c57eb44b177d91c70e1a0901/changelogs/current.yaml
This changed a few method signatures and will also require consumers update their extauth metric graphs.
In aggregate there are a few new metrics that consumors should be aware of in general as well.


chunter0:
Should we sync .bazelversion and .bazelrc to match [Envoy@1.25](https://github.com/envoyproxy/envoy/blob/c66619d13a1182848d8e3f69a304c4cce29673dd)?

nfuden (approved):

PR #204, opened 2023-02-17
transplant unwrap filter into open source

jbohanon:
The AWS Lambda feature `responseTransformation` is being modified to share the functionality for `unwrapAsApiGateway`. To facilitate this, the transformer that was previously implemented in `envoy-gloo-ee` has been moved here. 

Additionally, owing to the upgrade to envoy 1.25 where upstream moved the cluster implementations to be extensions, we need to build them explicitly.


source/extensions/transformers/aws_lambda/BUILD
```diff
@@ -0,0 +1,32 @@
...
+    repository = "@envoy",
+    deps = [
+        "//api/envoy/config/transformer/aws_lambda/v2:pkg_cc_proto",
+        "@envoy_gloo//source/extensions/filters/http/transformation:transformer_lib",
+        "@envoy//envoy/buffer:buffer_interface",
+        "@envoy//source/common/http:header_map_lib",
+        "@json//:json-lib",
+        "@envoy//source/common/common:minimal_logger_lib",
+        "@envoy//source/common/common:base64_lib",
+        "@envoy_gloo//source/extensions/filters/http/transformation:transformation_filter_config",
```
nfuden, line 28: Does this reference work? It should be local now right?

ben-taussig-solo:
This looks functional! One nit: I believe [this test](https://github.com/solo-io/envoy-gloo-ee/blob/master/test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc#L157-L201) in envoy-gloo-ee validates unwrap behavior -- could you add that to this PR if possible?

jbohanon:
> This looks good! One nit: I believe [this test](https://github.com/solo-io/envoy-gloo-ee/blob/master/test/extensions/filters/http/aws_lambda/aws_lambda_filter_test.cc#L157-L201) in envoy-gloo-ee validates unwrap behavior -- could you add that to this PR if possible?

It looks to me like that test is checking wrapAsApiGateway behavior? We are not configuring the unwrap transformer in the setup func.

ben-taussig-solo:
The updated test and ASan changes look good! I'll let OCTO approve but this looks good to me

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -0,0 +1,202 @@
...
+        ENVOY_STREAM_LOG(debug, "Api Gateway transformer cannot be used on the request path", stream_filter_callbacks);
+        return;
+      }
+      // we know we are on the repsonse path, so downcast RequestResponseHeaderMap to ResponseHeaderMap
+      Http::ResponseHeaderMap* response_headers = static_cast<Http::ResponseHeaderMap*>(&header_map);
+      try {
+        transform_response(response_headers, body, stream_filter_callbacks);
+      } catch (const std::exception &e) {
+        ENVOY_STREAM_LOG(debug, "Error transforming response: " + std::string(e.what()), stream_filter_callbacks);
+        ApiGatewayError error = ApiGatewayError{400, "400", "Failed to transform response"};
```
kdorosh, line 68: this feels more like a 5xx error than 4xx. the request was not malformed, it was a server error transforming the response

BUILD
```diff
@@ -16,17 +16,33 @@ envoy_cc_library(
...
+        "@envoy//source/extensions/clusters/aggregate:cluster",
+        "@envoy//source/extensions/clusters/dynamic_forward_proxy:cluster",
+        "@envoy//source/extensions/clusters/eds:eds_lib",
+        "@envoy//source/extensions/clusters/logical_dns:logical_dns_cluster_lib",
+        "@envoy//source/extensions/clusters/original_dst:original_dst_cluster_lib",
+        "@envoy//source/extensions/clusters/static:static_cluster_lib",
+        "@envoy//source/extensions/clusters/strict_dns:strict_dns_cluster_lib",
+    ],
+)
+
```
kdorosh, line 38: why do we need to add these?

api/envoy/config/transformer/aws_lambda/v2/api_gateway_transformer.proto
```diff
@@ -0,0 +1,12 @@
...
+package envoy.config.transformer.aws_lambda.v2;
+
+option java_package = "io.envoyproxy.envoy.config.transformer.aws_lambda.v2";
+option java_outer_classname = "ApiGatewayTransformerProto";
+option java_multiple_files = true;
+option go_package = "transformer";
+
+
+message ApiGatewayTransformation {
+}
```
kdorosh, line 12: why add an empty proto?

BUILD
jbohanon: I was receiving errors when running gloo that these extensions were not registered and found that upstream moved them out of common and made them extensions in 1.25. [Here is an example PR](https://github.com/envoyproxy/envoy/commit/1d60a116413a0422b2df50e5f6ef8b553caba53b) of one of the cluster types.

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+void ApiGatewayTransformer::transform(
+    Http::RequestOrResponseHeaderMap &header_map,
+    Http::RequestHeaderMap *request_headers,
+    Buffer::Instance &body,
+    Http::StreamFilterCallbacks &stream_filter_callbacks) const {
+      ENVOY_STREAM_LOG(debug, "Transforming response", stream_filter_callbacks);
+      // check if request header map equals response header map
+      if (*request_headers == header_map) {
```
kdorosh, line 57: is this just checking for pointer equality? shouldn't we be checking if the header sets are the same? or perhaps.. just don't even call this function on request path (envoy hooks are per request/response)

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -0,0 +1,202 @@
...
+  // all information about the request format is to be contained in the response body
+  // parse response body as JSON
+  const auto len = body.length();
+  const auto bodystring = absl::string_view(static_cast<char *>(body.linearize(len)), len);
+  nlohmann::json json_body;
+  try {
+    json_body = json::parse(bodystring);
+  } catch (std::exception& exception){
+    ENVOY_STREAM_LOG(debug, "Error parsing response body as JSON: " + std::string(exception.what()), stream_filter_callbacks);
+    ApiGatewayError error = {400, "400", "failed to parse response body as JSON"};
```
kdorosh, line 90: same nit about response code

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  // set response status code
+  if (json_body.contains("statusCode")) {
+    uint64_t status_value;
+    try {
+      status_value = json_body["statusCode"].get<uint64_t>();
+    } catch (std::exception& exception){
+      ENVOY_STREAM_LOG(debug, "Error parsing statusCode: " + std::string(exception.what()), stream_filter_callbacks);
+      ApiGatewayError error = {400, "400", "Non-integer status code"};
```
kdorosh, line 101: going to stop calling these out

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -0,0 +1,202 @@
...
+  // set response body
+  body.drain(body.length());
+  if (json_body.contains("body")) {
+    std::string body_dump;
+    if (json_body["body"].is_string()) {
+      body_dump = json_body["body"].get<std::string>();
+    } else {
+      body_dump = json_body["body"].dump();
+    }
+    if (json_body.contains("isBase64Encoded")) {
```
kdorosh, line 159: we should check the value of this key (true/false)

api/envoy/config/transformer/aws_lambda/v2/api_gateway_transformer.proto
jbohanon: As I understand it, this was needed to specify the transformer in typed config ([here](https://github.com/solo-io/gloo/pull/6762#discussion_r924705768) is a comment from the original implementation and [here](https://github.com/solo-io/envoy-gloo/blob/9fcbba0f96213546b91a97bcf99fd36009c31bb4/source/extensions/transformers/aws_lambda/api_gateway_transformer.h#L34-L36) is the function where we consume this, for registering the extension I believe)

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
jbohanon: I believe it is using pointer equality to determine if it's being called from the request or response. We only call [this](https://github.com/solo-io/envoy-gloo/blob/9fcbba0f96213546b91a97bcf99fd36009c31bb4/source/extensions/filters/http/aws_lambda/aws_lambda_filter.cc#L180) in the `encodeData` and `encodeTrailers` functions, so it is codified to not be on the request path. The only way I could see this being an issue is if the control plane introduced a bug to specify this transformer as the request transformer. In light of that, I think the defensiveness is warranted

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
jbohanon: in-lined the conditional [d809aac](https://github.com/solo-io/envoy-gloo/pull/204/commits/d809aac6ac333bb6af35b37a38aa5969f559d89f)

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
jbohanon: changed [d809aac](https://github.com/solo-io/envoy-gloo/pull/204/commits/d809aac6ac333bb6af35b37a38aa5969f559d89f)

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
kdorosh: I think we can make a new class `ResponseTransformer` (if does not already exist) to extend `Transformer` so we don't need this check and future devs don't need to reason about this potential misuse

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -0,0 +1,200 @@
...
+  // set response body
+  body.drain(body.length());
+  if (json_body.contains("body")) {
+    std::string body_dump;
+    if (json_body["body"].is_string()) {
+      body_dump = json_body["body"].get<std::string>();
+    } else {
+      body_dump = json_body["body"].dump();
+    }
+    if (json_body.contains("isBase64Encoded") && json_body["isBase64Encoded"] == true) {
```
kdorosh, line 159: we should check if `json_body["isBase64Encoded"]` is a bool type and then get / cast to bool before comparing

BUILD
```diff
@@ -16,17 +16,33 @@ envoy_cc_library(
...
 envoy_cc_binary(
     name = "envoy",
     repository = "@envoy",
     stamped = True,
     deps = [
         ":envoy_gloo_all_filters_lib",
+        ":envoy_all_clusters_lib",
```
kdorosh, line 45: we spoke about this offline, going to pull in upstream's version of this target

source/extensions/filters/http/transformation/body_header_transformer.h
```diff
@@ -17,10 +18,34 @@ class BodyHeaderTransformer : public Transformer {
                  Buffer::Instance &body,
                  Http::StreamFilterCallbacks &) const override;
   bool passthrough_body() const override { return false; };
+  //RequestTransformerConstSharedPtr asRequestTransformer();
```
kdorosh, line 21: remove?

source/extensions/filters/http/transformation/transformation_filter_config.cc
```diff
@@ -39,6 +40,77 @@ TransformerConstSharedPtr Transformation::getTransformer(
...
+  case envoy::api::v2::filter::http::Transformation::kTransformerConfig: {
+    auto &factory = Config::Utility::getAndCheckFactory<OnStreamCompleteTransformerExtensionFactory>(transformation.transformer_config());
+    auto config = Config::Utility::translateAnyToFactoryConfig(transformation.transformer_config().typed_config(), context.messageValidationContext().staticValidationVisitor(), factory);
+    return factory.createOnStreamCompleteTransformer(*config, context);
+  }
+  case envoy::api::v2::filter::http::Transformation::
+      TRANSFORMATION_TYPE_NOT_SET:
+    // TODO: return null here?
+  default:
+    throw EnvoyException("non existant transformation");
```
kdorosh, line 62: nit existent

source/extensions/filters/http/transformation/transformation_filter_config.cc
```diff
@@ -39,6 +40,77 @@ TransformerConstSharedPtr Transformation::getTransformer(
...
+    throw EnvoyException("invalid transformer configured on onstreamcomplete: HeaderBodyTransform");
+  }
+  case envoy::api::v2::filter::http::Transformation::kTransformerConfig: {
+    auto &factory = Config::Utility::getAndCheckFactory<OnStreamCompleteTransformerExtensionFactory>(transformation.transformer_config());
+    auto config = Config::Utility::translateAnyToFactoryConfig(transformation.transformer_config().typed_config(), context.messageValidationContext().staticValidationVisitor(), factory);
+    return factory.createOnStreamCompleteTransformer(*config, context);
+  }
+  case envoy::api::v2::filter::http::Transformation::
+      TRANSFORMATION_TYPE_NOT_SET:
+    // TODO: return null here?
```
kdorosh, line 60: I think this can be the same as default case

kdorosh, line 60: I think we should remove the TODO and add trace logs if that's easy

source/extensions/filters/http/transformation/transformer.h
```diff
@@ -53,32 +53,80 @@ class Transformer {
...
 
+class RequestTransformer : public Transformer {
+public:
+  virtual ~RequestTransformer() {}
+
+  virtual bool passthrough_body() const PURE;
+
+  virtual void transform(Http::RequestOrResponseHeaderMap &map,
+                         // request header map. this has the request header map
+                         // even when transforming responses.
```
kdorosh, line 64: i don't think this is accurate anymore? that was the whole point of creating separate classes?

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -0,0 +1,199 @@
...
+  if (json_body.contains("body")) {
+    std::string body_dump;
+    if (json_body["body"].is_string()) {
+      body_dump = json_body["body"].get<std::string>();
+    } else {
+      body_dump = json_body["body"].dump();
+    }
+    if (json_body.contains("isBase64Encoded")) {
+        auto is_base64 = json_body["isBase64Encoded"];
+      if (is_base64.is_boolean() && is_base64.get<bool>() == true) {
```
kdorosh, line 157: nit, spacing

test/extensions/filters/http/aws_lambda/aws_lambda_transformer_test.cc
```diff
@@ -60,19 +60,26 @@ class AWSLambdaTransformerTest : public testing::Test,
 protected:
   void SetUp() override { setupRoute(); }
 
-  const std::string TRANSFORMATION_YAML =
+  const std::string REQUEST_TRANSFORMATION_YAML =
   R"EOF(
-  name: io.solo.transformer.api_gateway_test_transformer
+  name: io.solo.requesttransformer.api_gateway_test_transformer
   typed_config:
-    "@type": type.googleapis.com/envoy.test.extensions.transformation.ApiGatewayTestTransformer
```
kdorosh, line 67: this is a break in backwards compatibility. even if deprecated, we should make sure the old format works for at least a release before removing so we have a zero downtime upgrade path. otherwise during upgrade if control plane upgrades and sends new APIs before dataplane updates then we can have dataplane downtime

kdorosh: if this does in fact still work, we should have the old test remain and just add new ones. and add appropriate deprecation warnings for removal in next Gloo Edge/Gloo platform release (we don't support jump upgrades)

jbohanon: The types of the actual transformers are not changing. The wrapAsApiGateway transformer (on the request path) is in EE and will be updated to be a `RequestTransformer` only. The unwrapAsApiGateway transformer (on the response path) is now in OSS and is a `ResponseTransformer` only.

The reason I needed to distinguish here is because the fake transformer used for testing was registered only once (as a `Transformer`). So to use a distinct `RequestTransformer` for request path and `ResponseTransformer` for response path, I had to make this distinction in the tests.

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -0,0 +1,199 @@
...
+    response_headers.setReferenceKey(AMAZON_ERRORTYPE_HEADER, error.code);
+    body.drain(body.length());
+    auto error_message = absl::StrCat(error.code, ": ", error.message);
+    body.add(error_message);
+    response_headers.setContentLength(body.length());
+  }
+
+void ApiGatewayTransformer::transform(
+    Http::RequestOrResponseHeaderMap &header_map,
+    Http::RequestHeaderMap *request_headers,
```
jbohanon, line 52: nit: remove this field name and below `(void)request_headers`

source/extensions/filters/http/transformation/transformer.h
jbohanon: good call, removed the comments where no longer relevant [8e3a137](https://github.com/solo-io/envoy-gloo/pull/204/commits/8e3a137861f451dea9a888c00cc9f9e2712b8a14)

source/extensions/filters/http/transformation/transformation_filter_config.cc
jbohanon: changed x4 [8e3a137](https://github.com/solo-io/envoy-gloo/pull/204/commits/8e3a137861f451dea9a888c00cc9f9e2712b8a14)

source/extensions/filters/http/transformation/transformation_filter_config.cc
jbohanon: fixed x4 [8e3a137](https://github.com/solo-io/envoy-gloo/pull/204/commits/8e3a137861f451dea9a888c00cc9f9e2712b8a14)

source/extensions/filters/http/transformation/body_header_transformer.h
jbohanon: removed [8e3a137](https://github.com/solo-io/envoy-gloo/pull/204/commits/8e3a137861f451dea9a888c00cc9f9e2712b8a14)

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
jbohanon: fixed [00fc6b3](https://github.com/solo-io/envoy-gloo/pull/204/commits/00fc6b3a8510c54a4f5bbc20ac6b7e7043e35fe3)

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
jbohanon: fixed [00fc6b3](https://github.com/solo-io/envoy-gloo/pull/204/commits/00fc6b3a8510c54a4f5bbc20ac6b7e7043e35fe3)

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -111,6 +111,47 @@ class InjaTransformer : public Transformer {
   bool merged_extractors_to_body_{};
 };
 
+class InjaRequestTransformer : public RequestTransformer, public InjaTransformer  {
+public:
+  InjaRequestTransformer(const envoy::api::v2::filter::http::TransformationTemplate
+                      &transformation);
+  ~InjaRequestTransformer(){};
+  void transform(Http::RequestOrResponseHeaderMap &map,
+                 Http::RequestHeaderMap *request_headers,
```
kdorosh, line 120: we shouldn't need to pass both? if this is a request transformer

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+void ApiGatewayTransformer::transform(
+    Http::RequestOrResponseHeaderMap &header_map,
+    Http::RequestHeaderMap *,
+    Buffer::Instance &body,
+    Http::StreamFilterCallbacks &stream_filter_callbacks) const {
+      ENVOY_STREAM_LOG(debug, "Transforming response", stream_filter_callbacks);
+      // we know we are on the repsonse path, so downcast RequestResponseHeaderMap to ResponseHeaderMap
+      Http::ResponseHeaderMap* response_headers = static_cast<Http::ResponseHeaderMap*>(&header_map);
```
kdorosh, line 57: I think this cast should be done before the interface receives the `header_map` when implementing the `transform()` function

jbohanon:
@kdorosh The refactoring changes proposed have been extracted into a [separate PR](https://github.com/solo-io/envoy-gloo/pull/214) for cleanliness and in order to not unnecessarily block this work from merging. It is expected to be a fast-follow, but the plan is to release envoy-gloo with just these changes in order to move forward with the lambda work, in particular to meet the Gloo Gateway 2.3 release. Is this agreeable?

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
jbohanon: This refactor is being worked on in a [separate PR](https://github.com/solo-io/envoy-gloo/pull/214)

ben-taussig-solo (commented):
Looks good to me, given that we're moving the `Transformation` refactor to another PR

kdorosh (approved):

PR #207, opened 2023-03-01
Add New Fields to requestTransformation output

ben-taussig-solo:
# Description
 - Adds the following new fields to the `bodyHeaderTransform` when `addRequestMetadata` is enabled:
    - `queryStringParameters`
    - `multiValueQueryStringParameters`
    - `multiValueHeaders`
 - When a user sends a header or query string parameter with the same key multiple times, the last-specifed value is used as the value of the parameter.
   - I.E., if a user sends the following request:
     ```
     GET /?foo=bar&foo=baz
     ```
     The value of `queryStringParameters["foo"]` will be `"baz"`.
     The value of `multiValueQueryStringParameters["foo"]` will be `["bar", "baz"]`.
   - This is the same behavior we use in our Enterprise Lambda request transformer (AKA `wrapAsApiGateway`)

# Technical Details
 - `parse_headers` and `parse_query_string` are adapted from portions of our [enterprise AWS Lambda Request Transformer](https://github.com/solo-io/envoy-gloo-ee/blob/master/source/extensions/transformers/aws_lambda_request_transformer/api_gateway_request_transformer.cc)
   - `parse_headers` was adapted from [this excerpt](https://github.com/solo-io/envoy-gloo-ee/blob/master/source/extensions/transformers/aws_lambda_request_transformer/api_gateway_request_transformer.cc#L51-L70)
   - `parse_query_string` was adapted from [this excerpt](https://github.com/solo-io/envoy-gloo-ee/blob/master/source/extensions/transformers/aws_lambda_request_transformer/api_gateway_request_transformer.cc#L71-L88)
 - `parse_parameters` is cribbed from [envoy-gloo-ee](https://github.com/solo-io/envoy-gloo-ee/blob/65b9648c49f555571a5b9baf5153d502e6b30345/source/extensions/transformers/aws_lambda_request_transformer/api_gateway_request_transformer.cc#L131-L158), which itself is a slight modification of an envoy utility to allow for the parsing of multi value query string parameters.
   - I think we should remove the `parse_parameters` function from envoy-gloo-ee and instead call into the one defined in this PR. 

# TODO
 - This PR is functional in its current state. However, some more changes are needed in other repos to make this feature complete.
## Gloo
 - [ ] e2e testing against the new functionality added in this PR
   - [ ] validate that the new fields are added to the request when `addRequestMetadata` is enabled
   - [ ] validate that the new functionality behaves as expected when leveraged by the AWS lambda filter
 - [ ] Documentation updates to reflect the new functionality added in this PR

## Envoy-gloo-ee
 - [ ] remove the `parse_parameters` function from envoy-gloo-ee and instead call into the one defined in this PR.


source/extensions/filters/http/transformation/body_header_transformer.cc
```diff
@@ -62,6 +67,94 @@ void BodyHeaderTransformer::transform(
...
+            if (headers[header_key].empty()) {
+                headers[header_key] = header_value;
+            } else {
+                // If there are more than one headers with the same key, use the last one
+                auto existing_header_value = headers[header_key];
+                headers[header_key] = header_value;
+ 
+                // Then, ensure all existing values are in the multiValueHeaders vector
+                if (multi_value_headers[header_key].empty()) {
+                    multi_value_headers[header_key].push_back(existing_header_value);
```
jbohanon, line 88: why do we use push_back here and emplace_back below? Is it situational, or is one more correct than the other?

ben-taussig-solo: A useful explanation for the difference b/w the two is here: https://www.quora.com/In-what-scenarios-should-I-use-emplace_back-instead-of-push_back-in-C%2B%2B/answer/Kevin-Mather?ch=10&share=46e9e37c&srid=CWXc

It seems like emplace_back is faster than push_back in some situations, so I'm choosing to use that in each location.

ashishb-solo:
Question regarding the implementation: if there are 2 instances of a header, there will be 2 entries for that header in `multiValueQueryStringParameters`. If there is only one instance of a header, then `multiValueQueryStringParameters` for that instance will not have any entries.

Is this the intended behaviour? Based on my reading of both the description and the code, that's what appears to be happening and i just wanted to make sure that's the intent.

ben-taussig-solo:
> Question regarding the implementation: if there are 2 instances of a header, there will be 2 entries for that header in `multiValueQueryStringParameters`. If there is only one instance of a header, then `multiValueQueryStringParameters` for that instance will not have any entries.
> 
> Is this the intended behaviour? Based on my reading of both the description and the code, that's what appears to be happening and i just wanted to make sure that's the intent.

Yes, this behavior matches with how AWS API gateway (and our `wrapAsApiGateway` feature) handles multi-value headers and queryString parameters. The details are:
 - If there is only one value for a header/queryString parameter, that header/param will not have entries in the multiValue object
 - If there is more than one value for a header/queryString parameter:
   - All values will be present in the relevant multiValue object
   - The last value assigned to the header/param will be listed as the value in the `headers`/`queryStringParameters` object

source/extensions/filters/http/transformation/body_header_transformer.cc
```diff
@@ -62,6 +67,97 @@ void BodyHeaderTransformer::transform(
...
+          multi_value_query_string_parameters[pair.first] = std::vector<std::string>();
+          multi_value_query_string_parameters[pair.first].emplace_back(existing_value);
+        }
+        multi_value_query_string_parameters[pair.first].emplace_back(pair.second);
+      }
+    }
+}
+
+// explain the differences between push_back and emplace_back:
+// the diff
```
kdorosh, line 127: this looks incomplete?

source/extensions/filters/http/transformation/body_header_transformer.cc
```diff
@@ -24,17 +24,16 @@ void BodyHeaderTransformer::transform(
...
-      [&headers](const Http::HeaderEntry &header) -> Http::HeaderMap::Iterate {
-        headers[std::string(header.key().getStringView())] =
-            std::string(header.value().getStringView());
-        return Http::HeaderMap::Iterate::Continue;
-      });
+  auto headers = std::map<std::string, absl::string_view>{};
+  // multi_value_headers will not be populated if add_request_metadata_ is false
+  auto multi_value_headers = std::map<std::string, std::vector<absl::string_view>>{};
+  parse_headers(header_map, headers, multi_value_headers);
+  json_body["headers"] = headers;
```
kdorosh, line 31: we assign `headers` to json_body` here, but `headers` only holds values as string views (which may or may not outlive `json_body`. i think this could be a bug, and it would be much safer imo to make `headers` and `multiValueHeaders` maps hold values that are strings instead of string views

source/extensions/filters/http/transformation/body_header_transformer.cc
```diff
@@ -24,17 +24,16 @@ void BodyHeaderTransformer::transform(
     json_body["body"] = body.toString();
   }
 
-  json &headers = json_body["headers"];
-  header_map.iterate(
-      [&headers](const Http::HeaderEntry &header) -> Http::HeaderMap::Iterate {
-        headers[std::string(header.key().getStringView())] =
-            std::string(header.value().getStringView());
```
kdorosh, line 31: see how the old code made new strings in the headers map? i think we should do the same when parsing headers and multi value headers

changelog/v1.25.1-patch3/request-transform-improvements.yaml
```diff
@@ -0,0 +1,9 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/solo-projects/issues/4571
+  resolvesIssue: false
+  description: >
+    Adds support for queryStringParameters, multiValueQueryStringParameters, and
+    multiValueHeaders to the requestTransformation feature in the transformation
+    filter. This feature is leveraged by the requestTransformation setting in our
+    AWS Lambda filter.
```
kdorosh, line 9: i think we should also update the proto comment to reflect the new fields added as well https://github.com/solo-io/envoy-gloo/blob/74cea33874f2bd0ec4f10c7590023c6d2ff53689/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto#L265-L267

changelog/v1.25.1-patch3/request-transform-improvements.yaml
```diff
@@ -0,0 +1,9 @@
+changelog:
+- type: NEW_FEATURE
+  issueLink: https://github.com/solo-io/solo-projects/issues/4571
+  resolvesIssue: false
+  description: >
```
kdorosh, line 5: super nit;
```suggestion
  description: >-
```

source/extensions/filters/http/transformation/body_header_transformer.cc
```diff
@@ -62,6 +67,97 @@ void BodyHeaderTransformer::transform(
...
+                headers[header_key] = header_value;
+ 
+                // Then, ensure all existing values are in the multiValueHeaders vector
+                if (multi_value_headers[header_key].empty()) {
+                    multi_value_headers[header_key].emplace_back(existing_header_value);
+                }
+
+                multi_value_headers[header_key].emplace_back(header_value);
+            }
+            return Http::HeaderMap::Iterate::Continue;
```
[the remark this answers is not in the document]
ben-taussig-solo: Did some cleanup here: https://github.com/solo-io/envoy-gloo/pull/207/commits/3081fa2a55ea0e22ead1787389c6f7452ff8770d

I don't think the proposed change quite works though; in order to match AWS's behavior and the `wrapAsApiGateway` implementation, multi_value_headers _only_ contains entries for headers which have multiple values. I.E., if a header only has one value, it should not have an entry in multi_value_headers

changelog/v1.25.1-patch3/request-transform-improvements.yaml
ben-taussig-solo: Updated here: https://github.com/solo-io/envoy-gloo/pull/207/commits/16a219dc3d7e41348504d55520de25a5e63a9e13

source/extensions/filters/http/transformation/body_header_transformer.cc
ben-taussig-solo: Done here: https://github.com/solo-io/envoy-gloo/pull/207/commits/922b07899c3bdbe007ef8e0692a1a02f42b6255d

source/extensions/filters/http/transformation/body_header_transformer.cc
ben-taussig-solo: Done here: https://github.com/solo-io/envoy-gloo/pull/207/commits/922b07899c3bdbe007ef8e0692a1a02f42b6255d ,
 with some additional changes here: https://github.com/solo-io/envoy-gloo/pull/207/commits/3081fa2a55ea0e22ead1787389c6f7452ff8770d

source/extensions/filters/http/transformation/body_header_transformer.cc
ben-taussig-solo: This was a note-to-self that I accidentally committed. I have removed the comment with this commit: https://github.com/solo-io/envoy-gloo/commit/922b07899c3bdbe007ef8e0692a1a02f42b6255d

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -262,7 +262,8 @@ message Passthrough {}
 message MergeExtractorsToBody {}
 
 message HeaderBodyTransform {
-  // When transforming a request, setting this to true will additionally add "queryString", 
-  // "httpMethod" and "path" to the body
+  // When transforming a request, setting this to true will additionally add
+  // "queryString", "queryStringParameters", "multiValueQueryStringParameters",
+  // "multiValueHeaders", "httpMethod" and "path" to the body
```
kdorosh, line 267: can we clarify that multi value is omitted if there's only one value (to match AWS behavior)?

kdorosh (approved):

PR #208, opened 2023-03-03
[v1.24.x] Add New Fields to requestTransformation output 

ben-taussig-solo:
# Description
 - Backport #207 to v1.24.x


jbohanon:
@ben-taussig-solo Is this PR still needed since we will be able to use the 1.25 release for this work?

PR #209, opened 2023-03-03
Address several asan fixes

ashishb-solo:
Most of the issues are bugs, but the one in the lambda filter has a small memory leak that is fixed here

### DIFF
--- a/test/integration/transformation_filter_integration_test.cc
+++ b/test/integration/transformation_filter_integration_test.cc
@@ -475,12 +475,13 @@ TEST_P(TransformationFilterIntegrationTest, BodyHeaderTransform) {
   EXPECT_EQ(expected_request, actual_request);
 
   json actual_response = json::parse(response->body());
+  // remove the `x-envoy-upstream-service-time` as it is not deterministic.
+  actual_response["headers"].erase("x-envoy-upstream-service-time");
   auto expected_response = R"( 
   {
     "headers": {
       ":status":"200",
-      "content-length":"0",
-      "x-envoy-upstream-service-time":"0"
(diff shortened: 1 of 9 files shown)


changelog/v1.25.1-patch2/asan-fixes.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+- type: NON_USER_FACING
+  issueLink: https://github.com/solo-io/envoy-gloo-ee/pull/481
+  resolvesIssue: false
+  description: Fix issues identified by ASAN
```
kdorosh, line 5: at least one is a memory leak, so this could be a fix?

test/integration/transformation_filter_integration_test.cc
```diff
@@ -475,12 +475,13 @@ TEST_P(TransformationFilterIntegrationTest, BodyHeaderTransform) {
   EXPECT_EQ(expected_request, actual_request);
 
   json actual_response = json::parse(response->body());
+  // remove the `x-envoy-upstream-service-time` as it is not deterministic.
```
kdorosh, line 478: can we be a bit more specific here? e.g. header order rendering is not deterministic

kdorosh (commented):
lgtm overall just nits!

test/integration/transformation_filter_integration_test.cc
ashishb-solo: https://github.com/solo-io/envoy-gloo/pull/209/commits/724f11cd3ba5b8a2849e120197e7ed97bad12373

kdorosh (approved):

PR #236, opened 2023-04-28
Fix for envoy patch application

ashishb-solo:
The patch values in repository_locations.bzl were actually being ignored. It
happens, the patch logic is a bit bizarre like that. In any case, this commit
rectifies it so the patches actually get applied.

### DIFF
--- a/bazel/foreign_cc/BUILD
+++ b/bazel/foreign_cc/BUILD
@@ -1,0 +1,1 @@
+licenses(["notice"])  # Apache 2
--- a/bazel/foreign_cc/BUILD
+++ b/bazel/foreign_cc/BUILD
@@ -1,1 +1,3 @@
+# this file can be left empty, but it is required in order to find the patches in bazel/repositories.bzl
+
 licenses(["notice"])  # Apache 2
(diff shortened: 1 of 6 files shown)


bazel/foreign_cc/BUILD
```diff
@@ -0,0 +1 @@
+licenses(["notice"])  # Apache 2
```
ben-taussig-solo, line 1: Did you intend to commit this?

ben-taussig-solo (commented):
Confirmed that:
 - All other patches have no `a/` or `b/` prefixes
 - All other patches are referenced in repositories.bzl, not respository_locations.bzl

So this looks good to me!

bazel/foreign_cc/BUILD
ashishb-solo: a `BUILD` file is required. i don't think the `licenses` line is necessary, but i don't think it's harmful either

ben-taussig-solo: Ah I see, I didn't realize there was no `BUILD` file in `bazel/foreign_cc` before this. So the reason that wasn't an issue before is that the patches weren't actually being consumed?

ashishb-solo: added [this comment](https://github.com/solo-io/envoy-gloo/pull/236/commits/e17f5cfe1391c1f14549ebe639eb0ce16b2cc248) to explain why the file is needed

ben-taussig-solo: Looks good, resolving this thread

ben-taussig-solo (approved):

PR #237, opened 2023-05-19
POC: Added replace_with_random() for substitutions in templates

dmitri-d:
A test transformation policy. Please note use of `replace_with_random` in `baz` header and in the response body:
```
kmgmt apply -f- <<EOF
apiVersion: trafficcontrol.policy.gloo.solo.io/v2
kind: TransformationPolicy
metadata:
  annotations:
    cluster.solo.io/cluster: ""
  name: transformation-1
  namespace: bookinfo
spec:
  applyToRoutes:
  - route:
      labels:
        route: ratings
  config:
    phase:
      postAuthz:
        priority: 10
    response:
      injaTemplate:
        advancedTemplates: false
        parseBodyBehavior: DontParse
        extractors:
          content-suffix:
            header: content-type
            regex: .*
        headers:
          baz:
            text: '{{ replace_with_random(content-suffix, "json") }}'          
          foo:
            text: bar
        body:
          text: ' {{ replace_with_random(body(), "Reviewer") }}'
EOF
```
An example without application of the transformation policy:
```
[wb@solo-system-ddolguikh proxy]$ curl -i -H"host: www.example.com" http://${INGRESS_GW_IP}:${INGRESS_PORT}/ratings/1
HTTP/1.1 200 OK
content-type: application/json
date: Fri, 19 May 2023 23:37:35 GMT
x-envoy-upstream-service-time: 0
server: istio-envoy
transfer-encoding: chunked

{"id":1,"ratings":{"Reviewer1":5,"Reviewer2":4}}
```

Same, but with the transformation policy applied. In `baz` header, the value of content-type header was used as the source string and the `json` part replaced with a random string. In the body `Reviewer` pattern was replaced with a random one.
```
[wb@solo-system-ddolguikh proxy]$ curl -i -H"host: www.example.com" http://${INGRESS_GW_IP}:${INGRESS_PORT}/ratings/1
HTTP/1.1 200 OK
content-type: application/json
date: Fri, 19 May 2023 23:33:09 GMT
x-envoy-upstream-service-time: 1
baz: application/1715106686
foo: bar
content-length: 53
server: istio-envoy

 {"id":1,"ratings":{"-6778386221":5,"-6778386222":4}}
```

Remaining questions:
 - Do we want to support multiple transformation policies per phase? Strictly speaking not necessary for this work, but there appears to be interest (see this thread: https://solo-io-corp.slack.com/archives/CHAC8587L/p1684511829499599), Support for multiple transformation policies per phase can probably be handled as a separate feature though.
 - what's the format for the replacement string? I don't think we want signed ints there though. Envoy's random number generator supports generation of uuids natively, if that's what we want to do.
 - `replace_with_random` function is a bit awkward, as it requires a source string as a parameter. Could implement it as a variadic-type function and populate a "special" var in the template context as a default source. For headers it would be the current header value, for the body -- the body itself.

### DIFF
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -2,6 +2,8 @@
 
 #include <iterator>
 
+#include "absl/strings/str_replace.h"
+
 #include "source/common/buffer/buffer_impl.h"
 #include "source/common/common/macros.h"
 #include "source/common/common/regex.h"
@@ -141,6 +143,9 @@ TransformerInstance::TransformerInstance(
   });
   env_.add_callback("substring", 3, [this](Arguments &args) {
     return substring_callback(args); 
+  });
+  env_.add_callback("replace_with_random", 2, [this](Arguments &args) {
+    return replace_with_random_callback(args);
   });
 }
 
@@ -311,6 +316,26 @@ json TransformerInstance::substring_callback(const inja::Arguments &args) const 
 
   // otherwise, return the substring from start to start + len
   return input.substr(start, substring_len);
+}
+
+json TransformerInstance::replace_with_random_callback(const inja::Arguments &args) {
+    // first argument: string to modify
+  const std::string &source = args.at(0)->get_ref<const std::string &>();
+    // second argument: pattern to be replaced
+  const std::string &to_replace = args.at(1)->get_ref<const std::string &>();
+
+  return 
+    absl::StrReplaceAll(source, {{to_replace, absl::StrCat(random_for_pattern(to_replace))}});
+}
+
+int TransformerInstance::random_for_pattern(const std::string& pattern) {
+  auto found = pattern_replacements_.find(pattern);
+  if (found == pattern_replacements_.end()) {
+    auto random = rng_.random();
+    pattern_replacements_.insert({pattern, random});
+    return random;
+  }
+  return found->second;
 }
 
 std::string TransformerInstance::render(const inja::Template &input) {
--- a/source/extensions/filters/http/transformation/inja_transformer.h
+++ b/source/extensions/filters/http/transformation/inja_transformer.h
@@ -6,6 +6,7 @@
 #include "envoy/http/header_map.h"
 
 #include "source/common/common/base64.h"
+#include "source/common/common/random_generator.h"
 
 #include "source/extensions/filters/http/transformation/transformer.h"
 
@@ -47,6 +48,8 @@ private:
   nlohmann::json base64_encode_callback(const inja::Arguments &args) const;
   nlohmann::json base64_decode_callback(const inja::Arguments &args) const;
   nlohmann::json substring_callback(const inja::Arguments &args) const;
+  nlohmann::json replace_with_random_callback(const inja::Arguments &args);
+  int random_for_pattern(const std::string& pattern);
 
   inja::Environment env_;
   const Http::RequestOrResponseHeaderMap &header_map_;
@@ -56,6 +59,8 @@ private:
   const nlohmann::json &context_;
   const std::unordered_map<std::string, std::string> &environ_;
   const envoy::config::core::v3::Metadata *cluster_metadata_;
+  std::unordered_map<std::string, int> pattern_replacements_;
+  Envoy::Random::RandomGeneratorImpl rng_;
 };
 
 class Extractor : Logger::Loggable<Logger::Id::filter> {
--- a/source/extensions/filters/http/transformation/BUILD
+++ b/source/extensions/filters/http/transformation/BUILD
@@ -91,6 +91,7 @@ envoy_cc_library(
         "@envoy//source/common/common:regex_lib",
         "@envoy//source/common/common:utility_lib",
         "@envoy//source/common/protobuf",
+	"@envoy//source/common/common:random_generator_lib",
         "@inja//:inja-lib",
         "@json//:json-lib",
     ],


dmitri-d:
Ping @marcogschmidt, @EItanya, @kdorosh, @saiskee, @yuval-k, @Sodman

kdorosh (changes requested):
i thought that we needed the same "nonce" type value everywhere? each call of replace with random here creates a new random value, ~and it doesn't seem easy for the latter argument to match a bunch of different sources for replace~ edit: cached per pattern so this is fine

also can we link the slab doc with the design to the pull request for easy reference?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -313,6 +318,26 @@ json TransformerInstance::substring_callback(const inja::Arguments &args) const
...
+  const std::string &to_replace = args.at(1)->get_ref<const std::string &>();
+
+  return 
+    absl::StrReplaceAll(source, {{to_replace, absl::StrCat(random_for_pattern(to_replace))}});
+}
+
+int TransformerInstance::random_for_pattern(const std::string& pattern) {
+  auto found = pattern_replacements_.find(pattern);
+  if (found == pattern_replacements_.end()) {
+    auto random = rng_.random();
```
kdorosh, line 334: any guarantees here about length and the chance of collisions? i think id slightly prefer the UID version of this func

kdorosh (commented):
i see the cache is per pattern so I think this will work, still would like to see tests for this callback

kdorosh (commented):
i see the cache is per pattern so I think this will work, still would like to see tests for this callback

dmitri-d:
> also can we link the slab doc with the design to the pull request for easy reference?

Added to PR description .

source/extensions/filters/http/transformation/inja_transformer.cc
dmitri-d: This uses envoy's `Envoy::Random::RandomGeneratorImpl`, which uses BoringSSL's `RAND_bytes` (billed as cryptographically secure pseudo rng, see here: https://www.openssl.org/docs/manmaster/man3/RAND_bytes.html) at its core.

As currently implemented it pre-fetches 256*64 bits of randomness, which it then returns as needed. Both `RandomGeneratorImpl::random()` and `RandomGeneratorImpl::uuid()` consume 64 bits of "randomness" at a time, a probability of collision is then 1/2<sup>64</sup>. 

[Link to `RandomGeneratorImpl`](https://github.com/envoyproxy/envoy/blob/main/source/common/common/random_generator.cc#L12).

dmitri-d:
* Updated replacement nonce format: it is now a 128 bit long random number, base64-encoded with no padding.
* Added tests

dmitri-d:
Same transformation policy as before and `replace_with_random()` with the latest changes:

```
[wb@solo-system-ddolguikh proxy]$ curl -i -H"host: www.example.com" http://${INGRESS_GW_IP}:${INGRESS_PORT}/ratings/1
HTTP/1.1 200 OK
content-type: application/json
date: Tue, 23 May 2023 23:00:20 GMT
x-envoy-upstream-service-time: 0
baz: application/ApNxE5S/PIkLDcmwFmhoQA
foo: bar
content-length: 77
server: istio-envoy

 {"id":1,"ratings":{"zsWQfWAvwOM0FgFvJ/Uraw1":5,"zsWQfWAvwOM0FgFvJ/Uraw2":4}}
```

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -936,6 +936,120 @@ TEST(Transformer, transformHeaderAndHeadersToRemove) {
...
+  auto formatted_string = fmt::format("{{{{replace_with_random(\"{}\", \"{}\")}}}}", test_string, pattern);
+
+  transformation.mutable_body()->set_text(formatted_string);
+
+  InjaTransformer transformer(transformation);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+
+  Buffer::OwnedImpl body("");
+  transformer.transform(headers, &headers, body, callbacks);
```
kdorosh, line 954: shouldn't the body here be nonempty and contain `test_string`? i must be missing something because the test presumably somehow passes and replaces the replace-me

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -936,6 +936,120 @@ TEST(Transformer, transformHeaderAndHeadersToRemove) {
...
+  auto test_string = "test-replace-me";
+  auto formatted_string = fmt::format("{{{{replace_with_random(\"{}\", \"{}\")}}}}", test_string, pattern);
+
+  transformation.mutable_body()->set_text(formatted_string);
+
+  InjaTransformer transformer(transformation);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+
+  Buffer::OwnedImpl body("");
```
kdorosh, line 953: i was expecting this to contain "test-replace-me" rather than setting it directly on the transformation mutable body

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -313,6 +318,31 @@ json TransformerInstance::substring_callback(const inja::Arguments &args) const
...
+
+std::string& TransformerInstance::random_for_pattern(const std::string& pattern) {
+  auto found = pattern_replacements_.find(pattern);
+  if (found == pattern_replacements_.end()) {
+    uint64_t random[3];
+    uint64_t high = rng_.random();
+    uint64_t low = rng_.random();
+    random[0] = low;
+    random[1] = high;
+    random[2] = 0;
```
kdorosh, line 339: why do we do this? can we at least include a comment to explain?

kdorosh (commented):
lgtm overall! just a question and some style nits

source/extensions/filters/http/transformation/inja_transformer.cc
dmitri-d: Added a comment. The gist is shipt wanted a 128bit long random number, base64-encoded, this is an implementation of that.

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -936,6 +936,120 @@ TEST(Transformer, transformHeaderAndHeadersToRemove) {
...
+  transformer.transform(headers, &headers, body, callbacks);
+  EXPECT_TRUE(body.toString().find("test-") != std::string::npos);
+  // length of "test-" + 128 bit long random number, Base64-encoded without padding (128/6 ~ 22).
+  EXPECT_EQ(27, body.toString().length());
+}
+
+TEST(InjaTransformer, ReplaceWithRandomHeaderTest) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
```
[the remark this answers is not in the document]
dmitri-d: Made a change similar to the search-and-replace test for the body.

source/extensions/filters/http/transformation/inja_transformer.cc
dmitri-d: Oh, and I removed the null at the end, as Base64::encode doesn't care if the string is null-terminated.

kdorosh (approved):

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -56,6 +59,8 @@ class TransformerInstance {
   const nlohmann::json &context_;
   const std::unordered_map<std::string, std::string> &environ_;
   const envoy::config::core::v3::Metadata *cluster_metadata_;
+  std::unordered_map<std::string, std::string> pattern_replacements_;
```
yuval-k, line 62: nit: in envoy `absl::flat_hash_map` is usually used instead of `std::unordered_map`

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -56,6 +59,8 @@ class TransformerInstance {
...
+  Envoy::Random::RandomGeneratorImpl rng_;
```
yuval-k, line 63: why is this created here directly? i see that it is usually coming from `context.api().randomGenerator()` ?
this object  will be re-initialized on every request, which makes me concerned that random maybe re-used if seed initialization is done based on time.

we have access to the context when the inja template is created here `source/extensions/filters/http/transformation/transformation_filter_config.cc:23`, so we can plumb the random generator through there if you agree that it's preferred

source/extensions/filters/http/transformation/inja_transformer.h
dmitri-d: nods, I saw we used `unordered_map` here and used the same for the consistency. I'll make the change.

PR #240, opened 2023-05-26
Expand Transformation Filter Logging

ben-taussig-solo:
# Description
 - Expand transformation logging so that request/response body and headers can be logged before a transformation is executed
   - More work to follow that will add similar functionality after transformations are executed
 - Adds new transformation-level setting `log_request_response_info` to enable this functionality
 - Adds new transformation-filter-level setting `log_request_response_info` to override the transformation-level setting

# Context
 - This PR is based off of this envoy-gloo-ee PR to expand logging for the GraphQL filter: https://github.com/solo-io/envoy-gloo-ee/pull/311/
 - I am struggling to write unit tests against this functionality. I have been trying to configure something like what is done in the upstream envoy `logs_handler_test.cc` (see [here](https://github.com/envoyproxy/envoy/blob/main/test/server/admin/logs_handler_test.cc#L32)), but have not been able to get a test against the content of a logger stood up yet
   - If any reviewers have experience/advice for standing up these sorts of tests, I would appreciate it.


api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -23,6 +23,10 @@ message FilterTransformations {
   // Only RouteTransformations.RouteTransformation with matching stage will be
   // used with this filter.
   uint32 stage = 2 [ (validate.rules).uint32 = {lte : 10} ];
+
+  // If set to true, the filter will log the request/response body and headers before 
+  // any transformation is applied.
+  bool log_request_response_info = 3;
```
kdorosh, line 29: this will not sanitize anything, yes? feels very similar in spirit to our closed source enterprise graphql logging API: https://github.com/solo-io/envoy-gloo-ee/blob/main/api/envoy/config/filter/http/graphql/v2/graphql.proto#L438-L440 we should be very explicit about the potential security risks of turning this on (logging aggregators persisting potentially sensitive information)

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -127,6 +131,10 @@ message Transformation {
     // Configuration for an externally implemented transformer.
     envoy.config.core.v3.TypedExtensionConfig transformer_config = 3;
   }
+
+  // If set to true, the filter will log the request/response body and headers before
+  // this transformation is applied.
+  bool log_request_response_info = 4;
```
kdorosh, line 137: this overrides any parent config, yes? can we document that?

source/extensions/filters/http/transformation/transformation_filter.cc
```diff
@@ -250,6 +251,12 @@ void TransformationFilter::transformSomething(
     void (TransformationFilter::*addData)(Buffer::Instance &)) {
 
   try {
+    // if log_request_response_info_ is set on the transformation, log the
+    // request body and request headers before transformation
+    TRANSFORMATION_SENSITIVE_LOG(debug, "headers before transformation: {}", 
+                          transformation, filter_config_, callbacks, header_map);
+    TRANSFORMATION_SENSITIVE_LOG(debug, "body before transformation: {}", 
+                          transformation, filter_config_, callbacks, body.toString());
     transformation->transform(header_map, request_headers_, body, callbacks);
```
kdorosh, line 260: for convenience, do you think we will want this after the transformation too? particularly for cases where we have staged / multiple transformations I could see this being useful

nfuden: I think the other extension is to log all Extracted values as well.

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
ben-taussig-solo: Yeah, this PR is pretty directly based off of those changes. I clarified the potential risks in this commit: https://github.com/solo-io/envoy-gloo/pull/240/commits/1bfb840c09c77fe810936e557e53f1cb4c54db44

```diff
+ // Logs request/response sensitive information
+ // By default, this is false so no request or response sensitive information is logged.
  // If set to true, the filter will log the request/response body and headers before 
  // any transformation is applied.
  bool log_request_response_info = 3;
```

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
ben-taussig-solo: Not quite -- The current conditional that I'm using here https://github.com/solo-io/envoy-gloo/blob/1bfb840c09c77fe810936e557e53f1cb4c54db44/source/extensions/filters/http/transformation/transformation_logger.h#L17-L20

basically boils down to 
```
if parent_setting || child_setting
```

|parent setting enabled|child setting enabled|logs emitted?|
|----------------------|----------------------|-------------|
|false|false|false|
|false|true|true|
|true|false|true|
|true|true|true|

The idea being that there's a filter-level toggle if you're trying to debug issues with a series of transformations, or a transformation-level toggle if you're trying to debug a single transformation -- how do you feel about this? I'm happy to change this behavior if you think there's a better alternative

ben-taussig-solo:
This is ready for another round of review. I'm leaving the work-in-progress tag on it, because I want to make sure the envoy-gloo-ee changes are ready to go before we merge this, given that any transformers defined in that repo will have to be updated

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -23,6 +24,12 @@ message FilterTransformations {
   // Only RouteTransformations.RouteTransformation with matching stage will be
   // used with this filter.
   uint32 stage = 2 [ (validate.rules).uint32 = {lte : 10} ];
+
+  // Logs request/response sensitive information
+  // By default, this is false so no request or response sensitive information is logged.
+  // If set to true, the filter will log the request/response body and headers before 
+  // any transformation is applied.
```
kdorosh, line 31: per my earlier comment, this will now also log after the transformation. please update the api to match

ben-taussig-solo: Resolved here: https://github.com/solo-io/envoy-gloo/pull/240/commits/a767130c44692fb7a593df256b4b07ce0abb4c96

source/extensions/filters/http/transformation/body_header_transformer.cc
```diff
@@ -13,7 +13,8 @@ namespace Extensions {
 namespace HttpFilters {
 namespace Transformation {
 
-BodyHeaderTransformer::BodyHeaderTransformer(bool add_request_metadata):add_request_metadata_(add_request_metadata){}
+BodyHeaderTransformer::BodyHeaderTransformer(bool add_request_metadata, google::protobuf::BoolValue log_request_response_info)
```
dmitri-d, line 16: A drive-by comment: using `google::protobuf::BoolValue` feels a bit weird here, why not use `bool` instead? This also complicates code downstream, as now a check must be made if log_request_response_info contains a value before checking the value itself. 

Another nit with this is that `log_request_response_info` is being passed by value, which can be expensive with protobufs, as it tries to hold on to any allocated memory, even after objects are cleared/destroyed.

ben-taussig-solo: Kevin suggested using `BoolValue` here https://github.com/solo-io/envoy-gloo/pull/240#discussion_r1207081565
Doing so allows the transformation-level setting to always take precedence over the listener-level setting, if defined

dmitri-d: Ty for explanation, that makes sense. A comment there might be useful to clear this up. Fwiw, `absl::optional<bool>` would work here, but at this point it would mostly be about not allocating lots of small protobuf objects.

kdorosh (approved):

source/extensions/filters/http/transformation/body_header_transformer.cc
ben-taussig-solo: This just merged -- I'm a little surprised because I thought the work-in-progress tag would prevent that -- I'm going to put up a follow-up PR that clarifies this point and passes around `log_request_response_info` by reference after I handle the necessary envoy-gloo-ee changes

PR #241, opened 2023-06-02
Update `create_transformer` signature to facilitate debug logging in enterprise transformations

ben-taussig-solo:
# Description
 - Addresses a blind spot missed in https://github.com/solo-io/envoy-gloo/pull/240
 - That PR added settings which could be used to enable additional debug logging for transformations processed via the transformation filter. However, it failed to support this feature in transformations not defined in envoy-gloo (i.e., enterprise transformations defined in envoy-gloo-ee)
 - This PR adds support for this functionality in transformations defined outside of this repository.

### DIFF
--- a/source/extensions/filters/http/transformation/transformation_filter_config.cc
+++ b/source/extensions/filters/http/transformation/transformation_filter_config.cc
@@ -29,7 +29,7 @@ TransformerConstSharedPtr Transformation::getTransformer(
   case envoy::api::v2::filter::http::Transformation::kTransformerConfig: {
     auto &factory = Config::Utility::getAndCheckFactory<TransformerExtensionFactory>(transformation.transformer_config());
     auto config = Config::Utility::translateAnyToFactoryConfig(transformation.transformer_config().typed_config(), context.messageValidationContext().staticValidationVisitor(), factory);
-    return factory.createTransformer(*config, context);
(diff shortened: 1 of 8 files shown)


source/extensions/filters/http/transformation/transformation_filter_config.cc
```diff
@@ -29,7 +29,7 @@ TransformerConstSharedPtr Transformation::getTransformer(
   case envoy::api::v2::filter::http::Transformation::kTransformerConfig: {
     auto &factory = Config::Utility::getAndCheckFactory<TransformerExtensionFactory>(transformation.transformer_config());
     auto config = Config::Utility::translateAnyToFactoryConfig(transformation.transformer_config().typed_config(), context.messageValidationContext().staticValidationVisitor(), factory);
-    return factory.createTransformer(*config, context);
+    return factory.createTransformer(*config, transformation.log_request_response_info(), context);
```
ben-taussig-solo, line 32: This line is called when we interact with a transformer defined outside of envoy-gloo. Without this change, that transformer will be created without knowledge of whether `log_request_response_info` has been set.

ben-taussig-solo: This is really the most important part of the PR -- the rest is just changing the signature of `createTransformer` and  updating the places where it's called to match that change

kdorosh (approved):

PR #243, opened 2023-06-08
[DNM] Bump Inja to latest version

jbohanon:
## Description
Upgrade Inja dependency to latest. The Inja library removed support for JSON Pointer syntax in v3.0. In order to keep this functionality available to our users, we have re-added via PR and are currently pointing to a commit hash in that PR branch from this repo.

## Context
This dependency upgrade involved a major rewrite of the Inja library in the v3.0 of the library. This rewrite caused us a breaking change, namely that it is now a parser error to attempt to parse templates containing callback functions which are not yet defined.

This transformer was originally structured with a separation of responsibilities: 
- `InjaTransformer` 
  - on construction
    - handles and stores all config data
    - parses certain input templates which validates them
    - stores them internally
  - on `transform`
    - does some additional template parsing which requires request data
    - constructs a `TransformerInstance`
    - handles the logic of what to render and where to put it 
- `TransformerInstance` 
  - registers custom callback functions
  - renders output

In order to enable parsing templates with the callback functions registered, the separation of responsibilities has shifted somewhat:
- `InjaTransformer` 
  - on construction
    - constructs an empty `TransformerInstance` to validate input templates
    - handles and stores some config data (much less than before)
  - on `transform`
    - handles additional config processing which requires request data
    - constructs a `TransformerInstance`
    - parses all input templates using the fully-hydrated instance
    - handles the logic of what to render and where to put it 
- `TransformerInstance` 
  - registers custom callback functions
  - parses input templates
  - renders output

A consequence of this restructuring is that we now parse certain templates twice (the templates that were previously parsed and stored in the `InjaTransformer` constructor). I have not yet evaluated what impact this has on performance.

### DIFF
--- a/bazel/repository_locations.bzl
+++ b/bazel/repository_locations.bzl
@@ -1,11 +1,11 @@
 REPOSITORY_LOCATIONS = dict(
     envoy = dict(
-        # envoy 1.26.2
-        commit = "4aa28dcbfdaf17b754ee6d4ca4f1fdf6f84c99dc",
+        # envoy 1.25.7
+        commit = "ddb94b9d031790282f87e51f7e59a9cd61e8e55b",
         remote = "https://github.com/envoyproxy/envoy",
     ),
     inja = dict(
-        commit = "4c0ee3a46c0bbb279b0849e5a659e52684a37a98",
+        commit = "5a18986825fc7e5d2916ff345c4369e4b6ea7122", # v3.4 + JSON Pointers
         remote = "https://github.com/pantor/inja",
     ),
     json = dict(
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -117,9 +117,6 @@ TransformerInstance::TransformerInstance(
     : header_map_(header_map), request_headers_(request_headers), body_(body),
       extractions_(extractions), context_(context), environ_(environ),
       cluster_metadata_(cluster_metadata), rng_(rng) {
-  for(auto it = extractions_.begin(); it != extractions_.end(); it++) {
-  }
-
   env_.add_callback("header", 1,
                     [this](Arguments &args) { return header_callback(args); });
   env_.add_callback("request_header", 1, [this](Arguments &args) {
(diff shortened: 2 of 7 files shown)


bazel/repository_locations.bzl
```diff
@@ -1,11 +1,11 @@
 REPOSITORY_LOCATIONS = dict(
     envoy = dict(
-        # envoy 1.26.2
-        commit = "4aa28dcbfdaf17b754ee6d4ca4f1fdf6f84c99dc",
+        # envoy 1.25.7
+        commit = "ddb94b9d031790282f87e51f7e59a9cd61e8e55b",
```
jbohanon, line 4:
```suggestion
        # envoy 1.26.2
        commit = "4aa28dcbfdaf17b754ee6d4ca4f1fdf6f84c99dc",
```

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -355,114 +357,127 @@ std::string TransformerInstance::render(const inja::Template &input) {
...
+  return env_.parse(input);
+}
+
+TransformerInstance TransformerInstance::empty_transformer_instance() {
+
+  std::unique_ptr<Http::RequestHeaderMap> emptyRequestHeaderMap;
+  emptyRequestHeaderMap = Http::RequestHeaderMapImpl::create();
+  auto header_map = emptyRequestHeaderMap.get();
+
+  absl::optional<std::string> string_body;
```
ben-taussig-solo, line 370: Does this get used?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -355,114 +357,127 @@ std::string TransformerInstance::render(const inja::Template &input) {
...
+  };
+
+  Buffer::OwnedImpl emptyBody{};
+  const envoy::config::core::v3::Metadata *cluster_metadata{};
+  json json_body;
+  std::unordered_map<std::string, absl::string_view> extractions;
+  Random::RandomGeneratorImpl rng;
+  std::unordered_map<std::string, std::string> empty_environ;
+
+  TransformerInstance instance(*header_map, header_map, get_body,
```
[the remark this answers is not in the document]
jbohanon: Good call; I reordered the definitions

source/extensions/filters/http/transformation/inja_transformer.cc
jbohanon: Nope! Good catch; removed

yuval-k (changes requested):
I mentioned this on slack, but for reference - i think we can use the old structure where most things are initialized in the ctor. To make this work with the new inja, we'll need to use envoy's ThreadLocal to temporary store the per-request template context ; that way the callbacks would be registered only once, at ctor time. and will get the different context via the TLS.

jbohanon:
Closing in favor of https://github.com/solo-io/envoy-gloo/pull/248

PR #248, opened 2023-06-23
Bump Inja dependency

jbohanon:
Uses thread-local storage to pass request data into the `TransformerInstance` so that we can use the instance (and its registered callbacks) to parse templates at init time.

### DIFF
--- a/bazel/repository_locations.bzl
+++ b/bazel/repository_locations.bzl
@@ -5,7 +5,7 @@ REPOSITORY_LOCATIONS = dict(
         remote = "https://github.com/envoyproxy/envoy",
     ),
     inja = dict(
-        commit = "4c0ee3a46c0bbb279b0849e5a659e52684a37a98",
+        commit = "5a18986825fc7e5d2916ff345c4369e4b6ea7122", # v3.4 + JSON Pointers
         remote = "https://github.com/pantor/inja",
     ),
     json = dict(
--- a/bazel/repository_locations.bzl
+++ b/bazel/repository_locations.bzl
@@ -6,7 +6,7 @@ REPOSITORY_LOCATIONS = dict(
     ),
     inja = dict(
         commit = "5a18986825fc7e5d2916ff345c4369e4b6ea7122", # v3.4 + JSON Pointers
-        remote = "https://github.com/pantor/inja",
+        remote = "https://github.com/solo-io/inja", # solo-io fork including the changes
     ),
     json = dict(
         commit = "bc889afb4c5bf1c0d8ee29ef35eaaf4c8bef8a5d",  # v3.11.2
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -343,6 +343,10 @@ std::string& TransformerInstance::random_for_pattern(const std::string& pattern)
   return found->second;
 }
 
+// parse calls Inja::Environment::parse which uses non-const references to member
+// data fields. This method is NOT SAFE to call outside of the InjaTransformer
+// constructor since doing so could cause Inja::Environment member fields to be
+// modified by multiple threads at runtime.
 inja::Template TransformerInstance::parse(std::string_view input) {
     return env_.parse(input);
 }
@@ -471,7 +475,8 @@ InjaTransformer::InjaTransformer(const TransformationTemplate &transformation,
 InjaTransformer::~InjaTransformer() {}
 
 // transform is called on the request path, and may be executed on any worker thread.
-// it must be thread-safe
+// it must be thread-safe. note that calling instance_->parse is NOT THREAD SAFE
+// and MUST NOT be done from this method.
 void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
                                 Http::RequestHeaderMap *request_headers,
                                 Buffer::Instance &body,
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -347,7 +347,7 @@ inja::Template TransformerInstance::parse(std::string_view input) {
     return env_.parse(input);
 }
 
-std::string TransformerInstance::render(const inja::Template &input) const {
+std::string TransformerInstance::render(const inja::Template &input) {
   // inja can't handle context that are not objects correctly, so we give it an
   // empty object in that case
   const auto& ctx = tls_.getTyped<ThreadLocalTransformerContext>();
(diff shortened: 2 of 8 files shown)


source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -53,15 +73,8 @@ class TransformerInstance {
...
   inja::Environment env_;
-  const Http::RequestOrResponseHeaderMap &header_map_;
-  const Http::RequestHeaderMap *request_headers_;
-  GetBodyFunc &body_;
-  const std::unordered_map<std::string, absl::string_view> &extractions_;
-  const nlohmann::json &context_;
-  const std::unordered_map<std::string, std::string> &environ_;
-  const envoy::config::core::v3::Metadata *cluster_metadata_;
   absl::flat_hash_map<std::string, std::string> pattern_replacements_;
-  Envoy::Random::RandomGenerator &rng_;
```
jbohanon, line 64: `rng_` can stay as a private field and be initialized during construction of the `InjaTransformer` since it is passed into that constructor at init time

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -526,17 +538,22 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
   if (ci.get()) {
     cluster_metadata = &ci->metadata();
   }
-
-  // start transforming!
-  TransformerInstance instance(header_map, request_headers, get_body,
-                               extractions, json_body, environ_,
-                               cluster_metadata, rng_);
-
+  tls_->set([header_map=&header_map, request_headers, get_body, extractions=&extractions, json_body=&json_body, environ=&environ_, cluster_metadata, rng=&rng_](Event::Dispatcher&) -> ThreadLocal::ThreadLocalObjectSharedPtr {
```
yuval-k, line 541: you can't call `set()` from `transform()` as `set` will set the object on all worker threads.
call `set()` from the ctor (you only need to call it one time).
Here you can call `getTyped` and update the existing object

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -24,18 +24,42 @@ namespace Transformation {
...
+      Http::RequestHeaderMap *request_headers,
+      GetBodyFunc *body,
+      std::unordered_map<std::string, absl::string_view> *extractions,
+      nlohmann::json *context,
+      std::unordered_map<std::string, std::string> *environ,
+      envoy::config::core::v3::Metadata *cluster_metadata)
+      : header_map_(header_map), request_headers_(request_headers), body_(body),
+      extractions_(extractions), context_(context), environ_(environ),
+      cluster_metadata_(cluster_metadata) {}
+
```
jbohanon, line 41: This constructor can be removed
```suggestion
```

jbohanon (commented):
I still want to add tests for 
- JSON pointer syntax
- short circuit evaluation
- too few arguments to function is a parser error

source/extensions/filters/http/transformation/inja_transformer.cc
jbohanon: I changed to have a `set()` call construct an empty object, then it is populated with request context by getting a reference to the thread-local object with `getTyped`. Is this more in line with how this should be used?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -104,17 +104,10 @@ Extractor::extractValue(Http::StreamFilterCallbacks &callbacks,
...
-    const Http::RequestHeaderMap *request_headers, GetBodyFunc &body,
-    const std::unordered_map<std::string, absl::string_view> &extractions,
-    const json &context,
-    const std::unordered_map<std::string, std::string> &environ,
-    const envoy::config::core::v3::Metadata *cluster_metadata,
-    Envoy::Random::RandomGenerator &rng)
-    : header_map_(header_map), request_headers_(request_headers), body_(body),
-      extractions_(extractions), context_(context), environ_(environ),
-      cluster_metadata_(cluster_metadata), rng_(rng) {
+// A TransformerInstance is constructed by the InjaTransformer::transform method
```
ben-taussig-solo, line 107: Is this still true? It seems like it's created by the injaTransformer constructor now

jbohanon: Good catch, thanks!

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -115,7 +122,8 @@ class InjaTransformer : public Transformer {
 
   absl::optional<inja::Template> body_template_;
   bool merged_extractors_to_body_{};
-  Envoy::Random::RandomGenerator &rng_;
+  ThreadLocal::SlotPtr tls_;
+  std::unique_ptr<TransformerInstance> instance_;
```
yuval-k, line 126: micro-nit: i don't think this needs to be a pointer
```suggestion
  TransformerInstance instance_;
```
(the .cc will change a bit as well)

bazel/repository_locations.bzl
```diff
@@ -5,7 +5,7 @@ REPOSITORY_LOCATIONS = dict(
         remote = "https://github.com/envoyproxy/envoy",
     ),
     inja = dict(
-        commit = "4c0ee3a46c0bbb279b0849e5a659e52684a37a98",
+        commit = "5a18986825fc7e5d2916ff345c4369e4b6ea7122", # v3.4 + JSON Pointers
         remote = "https://github.com/pantor/inja",
```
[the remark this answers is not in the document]
jbohanon: Good call, I made a solo-io fork, created a tag, and pointed to the commit in that repo. I should probably also update the source of the PR to be the solo-io branch

source/extensions/filters/http/transformation/inja_transformer.h
jbohanon: I followed the changes for this all the way until it needs a member function (`Environment::render`) within the Inja library to be marked `const`. I think this is because we're using it within the const `transform` method?

jbohanon: It's unclear why the build errors occur when this field is a concrete object, but not when it is a pointer. However, the performance overhead should be negligent. We confirmed that the `render` calls are thread safe because the `Renderer` [constructor](https://github.com/pantor/inja/blob/master/single_include/inja/inja.hpp#L2679) that ends up being called by `Inja::Environment::render` takes `const` parameters. The same cannot be said of the `Parser` [constructor](https://github.com/pantor/inja/blob/master/single_include/inja/inja.hpp#L2031), but we should not be calling `parse` on the request path anyway.

yuval-k (approved):

PR #258, opened 2023-07-19
Inja transformation preserve escaped json

jbohanon:
Users have the issue that when they transform a request or response body that includes escaped characters, for example this json request/response body 
```json
{
  "foo": "{\"bar\":\"baz\"}"
}
```

The characters lose their escaping when used in a template e.g. the template `{"response":"{{ foo }}"}` would produce the string 
```json
{"response":"{"bar":"baz"}"}
```

which is not valid JSON since the quotes are not escaped. Instead we expect to be able to produce
```json
{"response":"{\"bar\":\"baz\"}"}
```

This PR provides 2 methods of achieving this
- The new callback `raw_string` can be used to escape a particular value within a template.
  - e.g. the template would look like `{"response":"{{ raw_string(foo) }}"}`
- The transformation itself can be configured to do this for all string values found in the body and rendered in the template.
  - This is the new setting `render_body_as_json` on the `TransformationTemplate` setting

### DIFF
--- a/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
+++ b/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
@@ -257,7 +257,7 @@ message TransformationTemplate {
   repeated DynamicMetadataValue dynamic_metadata_values = 9;
 
   // Use to render the output of a body transformation as JSON. This will cause
-  // rendered values to be escaped in order to make valid JSON strings
+  // rendered string values to be escaped in order to make valid JSON strings
   bool render_body_as_json = 12;
 
 }
--- a/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
+++ b/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
@@ -256,9 +256,9 @@ message TransformationTemplate {
   // Use this field to set Dynamic Metadata.
   repeated DynamicMetadataValue dynamic_metadata_values = 9;
 
-  // Use to render the output of a body transformation as JSON. This will cause
-  // rendered string values to be escaped in order to make valid JSON strings
-  bool render_body_as_json = 12;
+  // Use to escape the output of a body transformation. This will cause
+  // rendered string values to be escaped in order to make valid JSON/YAML strings
+  bool escape_characters = 12;
 
 }
 
--- a/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
+++ b/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
@@ -204,9 +204,9 @@ message TransformationTemplate {
     InjaTemplate value = 2;
   }
 
-  // Use this attribute to transform request/response headers. It consists of 
-  // an array of string/template objects. Use this attribute to define multiple 
-  // templates for a single header. Header template(s) defined here will be appended to any 
+  // Use this attribute to transform request/response headers. It consists of
+  // an array of string/template objects. Use this attribute to define multiple
+  // templates for a single header. Header template(s) defined here will be appended to any
   // existing headers with the same header name, not replace existing ones.
   repeated HeaderToAppend headers_to_append = 10;
 
@@ -255,6 +255,11 @@ message TransformationTemplate {
   }
   // Use this field to set Dynamic Metadata.
   repeated DynamicMetadataValue dynamic_metadata_values = 9;
+
+  // Use to render the output of a body transformation as JSON. This will cause
+  // rendered values to be escaped in order to make valid JSON strings
+  bool render_body_as_json = 12;
+
 }
 
 // Defines an [Inja template](https://github.com/pantor/inja) that will be
--- a/bazel/repository_locations.bzl
+++ b/bazel/repository_locations.bzl
@@ -5,7 +5,10 @@ REPOSITORY_LOCATIONS = dict(
         remote = "https://github.com/envoyproxy/envoy",
     ),
     inja = dict(
-        commit = "5a18986825fc7e5d2916ff345c4369e4b6ea7122", # v3.4 + JSON Pointers
+        # Includes unmerged modifications for
+        # - JSON pointer syntax support
+        # - Allowing escaped strings
+        commit = "3aa95b8b58a525f86f79cb547bf937176c9cc7ff", # v3.4.0-patch1
         remote = "https://github.com/solo-io/inja", # solo-io fork including the changes
     ),
     json = dict(
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -143,6 +143,9 @@ TransformerInstance::TransformerInstance(ThreadLocal::Slot &tls, Envoy::Random::
   env_.add_callback("replace_with_random", 2, [this](Arguments &args) {
     return replace_with_random_callback(args);
   });
+  env_.add_callback("to_json", 1, [this](Arguments &args) {
+    return tojson_callback(args);
+  });
 }
 
 json TransformerInstance::header_callback(const inja::Arguments &args) const {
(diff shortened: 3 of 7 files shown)


api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -255,6 +255,11 @@ message TransformationTemplate {
   }
   // Use this field to set Dynamic Metadata.
   repeated DynamicMetadataValue dynamic_metadata_values = 9;
+
+  // Use to render the output of a body transformation as JSON. This will cause
+  // rendered values to be escaped in order to make valid JSON strings
```
jbohanon, line 260: nit:
```suggestion
  // rendered string values to be escaped in order to make valid JSON strings
```

bazel/repository_locations.bzl
```diff
@@ -5,7 +5,10 @@ REPOSITORY_LOCATIONS = dict(
         remote = "https://github.com/envoyproxy/envoy",
     ),
     inja = dict(
-        commit = "5a18986825fc7e5d2916ff345c4369e4b6ea7122", # v3.4 + JSON Pointers
+        # Includes unmerged modifications for
+        # - JSON pointer syntax support
+        # - Allowing escaped strings
+        commit = "3aa95b8b58a525f86f79cb547bf937176c9cc7ff", # v3.4.0-patch1
```
jbohanon, line 11: Before merging this we should get the [fork](https://github.com/solo-io/inja) sorted out

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -1147,6 +1147,70 @@ TEST_F(InjaTransformerTest, ParseUsingJsonPointerSyntax) {
...
+  TransformationTemplate transformation;
+  transformation.mutable_body()->set_text(
+      R"EOF({"Value":"{{ raw_string(value) }}"})EOF");
+  transformation.set_render_body_as_json(true);
+  InjaTransformer transformer(transformation, rng_, google::protobuf::BoolValue(), tls_);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+
+  Buffer::OwnedImpl body(R"({"value":"\"foo\""})"_json.dump());
+  auto expected_body = R"({"Value":"\\\"foo\\\""})"_json.dump();
```
jbohanon, line 1209: this is the expected double-escaping caused by using the callback `raw_string` when the `render_body_as_json` is also set to true for the transformation

test/integration/transformation_filter_integration_test.cc
```diff
@@ -493,4 +493,93 @@ TEST_P(TransformationFilterIntegrationTest, BodyHeaderTransform) {
...
+
+  auto actual_request_body = json::parse(upstream_request_->body().toString());
+  auto expected_request_body = R"({"Foo":"\"bar\""})"_json;
+
+  // make sure response works as well, with no metadata set:
+  EXPECT_EQ(expected_request_body, actual_request_body);
+
+  json actual_response_body = json::parse(response->body());
+  // remove the `x-envoy-upstream-service-time` since its
+  // value depends on how long the test took to run
```
jbohanon, line 536: these copypasta comments need to be removed

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -47,6 +47,9 @@ class TransformerInstance {
   void set_element_notation(inja::ElementNotation notation) {
       env_.set_element_notation(notation);
   };
+  void set_escape_strings(bool escape_strings) {
```
ben-taussig-solo, line 50: I think a comment here could be useful. I'm thinking about something along the lines of:
`// Sets the config for rendering strings raw or escaped`

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -255,6 +255,11 @@ message TransformationTemplate {
   }
   // Use this field to set Dynamic Metadata.
   repeated DynamicMetadataValue dynamic_metadata_values = 9;
+
+  // Use to render the output of a body transformation as JSON. This will cause
+  // rendered string values to be escaped in order to make valid JSON strings
+  bool render_body_as_json = 12;
```
kdorosh, line 261: i think we should rename this. perhaps this has similar utility for folks who want the body rendered for, say, yaml

kdorosh: say `escape_characters`

kdorosh (commented):
lgtm beyond naming nit

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -343,6 +346,32 @@ std::string& TransformerInstance::random_for_pattern(const std::string& pattern)
   return found->second;
 }
 
+json TransformerInstance::raw_string_callback(const inja::Arguments &args) const {
+  // inja::Arguments is a std::vector<const json *>, so we can get the json
+  // value from the args directly.
+  const auto& input = args.at(0);
```
nfuden, line 352: do we have a guarantee that this has non 0 length? If so can we add a link?

kdorosh: i think its line 146 parameter

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -343,6 +346,32 @@ std::string& TransformerInstance::random_for_pattern(const std::string& pattern)
...
+  // context (body) to be placed in exactly as-is. HOWEVER, the behavior of the jinja
+  // filter is such that the quotes added by .dumps() are left in. For that reason,
+  // this callback is NOT named to_json to avoid confusion with that behavior.
+
+  // strip the leading and trailing " characters that are added by dump()
+  // if C++20 is adopted, val.starts_with and val.ends_with would clean this up a bit
+  val = val.substr(0,1) == "\"" && val.substr(val.length()-1,1) == "\""
+      ? val.substr(1, val.length()-2)
+      : val;
+  return val;
```
kdorosh, line 372: did we see https://github.com/nlohmann/json/issues/1479#issuecomment-462900019

why is this preferred to their recommended solution? what about nested json that it will escape again?

source/extensions/filters/http/transformation/inja_transformer.cc
jbohanon: Yeah it's registered as requiring 1 argument. Inja will throw a parse exception if it's not precisely 1. I will find the relevant code and link it

source/extensions/filters/http/transformation/inja_transformer.cc
kdorosh: id like to see some tests with nested json; i suspect if we did that we would escape each child (we may not want) rather than just the outermost. but perhaps im wrong

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -1147,6 +1147,70 @@ TEST_F(InjaTransformerTest, ParseUsingJsonPointerSyntax) {
...
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+  TransformationTemplate transformation;
+  transformation.mutable_body()->set_text(
+      R"EOF({"Value":"{{ value }}"})EOF");
+  transformation.set_render_body_as_json(true);
+  InjaTransformer transformer(transformation, rng_, google::protobuf::BoolValue(), tls_);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+
+  Buffer::OwnedImpl body(R"({"value":"\"foo\""})"_json.dump());
```
kdorosh, line 1160: let's add some tests with nested json

source/extensions/filters/http/transformation/inja_transformer.cc
jbohanon: If the value was escaped as a string like so
```json
{"foo": "{\"bar\":\"baz\"}"}
```
then I would expect that the most atomic node that inja would parse would be to parse the "{\"bar\":\"baz\"}" as a string and _not_ then attempt to parse it as an object. That said, it's definitely a good question and worth testing

source/extensions/filters/http/transformation/inja_transformer.h
jbohanon: good call; added! 46f2a5f2597f208270d96a91568d61c2c17550a0

source/extensions/filters/http/transformation/inja_transformer.cc
jbohanon: added a comment with a link, as well as some unit tests. this is helpful because inja has an option to suppress exceptions, so I'm not sure what the behavior would be if that were to be turned on. In any event the test should fail if it is
46f2a5f2597f208270d96a91568d61c2c17550a0

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
jbohanon: I like it. I renamed the field

source/extensions/filters/http/transformation/inja_transformer.cc
jbohanon: I have added some tests with nested json to confirm the behavior. When valid JSON is fed in, regardless how much nesting in strings is done, there will not be double escaping 46f2a5f2597f208270d96a91568d61c2c17550a0

test/extensions/filters/http/transformation/inja_transformer_test.cc
jbohanon: added

source/extensions/filters/http/transformation/inja_transformer.cc
kdorosh: thank you, fantastic

kdorosh (approved):
lgtm

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -343,6 +346,34 @@ std::string& TransformerInstance::random_for_pattern(const std::string& pattern)
...
+  // because Inja will throw a Parser error in any other case.
+  // https://github.com/pantor/inja/blob/v3.4.0/include/inja/parser.hpp#L228-L231
+  const auto& input = args.at(0);
+
+  // make sure to bail if we're not working with a raw string value
+  if(!input->is_string()) {
+      return input->get_ref<const std::string&>();
+  }
+
+  auto val = input->dump();
```
kdorosh, line 361: https://github.com/solo-io/inja/pull/5#discussion_r1271046663

kdorosh (changes requested):
one last q per https://github.com/solo-io/envoy-gloo/pull/258/files#r1271047429

kdorosh (approved):

PR #263, opened 2023-08-08
eitanya/bump-release-1.27

EItanya:
### Description
In this PR we move our dependency on upstream Envoy up to 1.27.0. In doing this, we address a number of issues:

#### `do_ci.sh` overhaul
- Remove workaround where we checkout and old release of Envoy to get an old ci script
  - Add `distribution/` targets to build binary in mirror fashion to upstream
  - Add `contrib/` to suppress error
- Remove as many hacks as possible
- Deflate `zstd` using upstream's tool to access built binary

#### Python Root issue
Cloudbuild requires a container be run as root for its setup, but the latest python dependency within bazel forbids running as root. The dependency can be set up with a boolean flag `ignore_root_user_error` to suppress this error, which was added to the `WORKSPACE` file.

#### Regex
TODO

#### Static analysis
Static analysis began to fail on issues that it previously hadn't with our dependencies. We have separated this into a separate action in cloudbuild and marked it as not required.


ci/cleanup_do_ci.sh
```diff
@@ -0,0 +1,10 @@
+rm -rf \
```
nfuden, line 1: non-blocking given urgency. I propose in our follow up thread we also update our make targets to have some subsection of this

nfuden (approved):

ci/cleanup_do_ci.sh
jbohanon: I think they are all overwritten given our use of `cp -f` but I mainly added this to have for local dev. I also added the files to `.gitignore` but I think this still has value

kdorosh:
static analysis seems like something that will save us time in the long run as a required ci check

much quicker than debugging and dealing with later in customer envs

PR #266, opened 2023-08-18
[DNM] Hardcode us-east-2 when assuming new role

ben-taussig-solo:
# Description
 - Do not merge this PR. This is only created to generate a test build
 - The build will be used to validate whether the modified line is responsible for the issues described in this issue: https://github.com/solo-io/gloo/issues/8578
 - A subsequent PR will plumb a setting through so this value can be configured by an end-user


ben-taussig-solo:
/kick build failure, running again to see if it's transient
https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=7c09d68d-ed6b-424a-a6bb-9d43ff8e3998
`io.netty.channel.StacklessClosedChannelException`

PR #271, opened 2023-09-18
Use `do_ci.sh` from tagged upstream dependency

jbohanon:


ci/do_ci.sh
```diff
@@ -51,21 +48,17 @@ export BAZEL_EXTRA_TEST_OPTIONS="--test_env=ENVOY_IP_TEST_VERSIONS=v4only --test
...
+echo "test $BUILD_CONFIG" >> "${SOURCE_DIR}/test.bazelrc"
 
 echo Building
-sed -i "s|//test/tools/schema_validator:schema_validator_tool|@envoy//test/tools/schema_validator:schema_validator_tool|" "$UPSTREAM_ENVOY_SRCDIR/ci/do_ci.sh"
-sed -i "s|bazel-bin/test/tools/schema_validator/schema_validator_tool|bazel-bin/external/envoy/test/tools/schema_validator/schema_validator_tool|" "$UPSTREAM_ENVOY_SRCDIR/ci/do_ci.sh"
-sed -i "s|VERSION.txt|ci/FAKEVERSION.txt|" "$UPSTREAM_ENVOY_SRCDIR/ci/do_ci.sh"
-sed -i "s|\${ENVOY_SRCDIR}/VERSION.txt|ci/FAKEVERSION.txt|" "$UPSTREAM_ENVOY_SRCDIR/ci/build_setup.sh"
+sed -i 's|"//contrib/..."||' "$UPSTREAM_ENVOY_SRCDIR/ci/do_ci.sh"
+sed -i 's|//distribution|@envoy//distribution|' "$UPSTREAM_ENVOY_SRCDIR/ci/do_ci.sh"
+sed -i 's|bazel-bin/distribution|bazel-bin/external/envoy/distribution|' "$UPSTREAM_ENVOY_SRCDIR/ci/do_ci.sh"
```
jbohanon, line 61: It's possible that this causes envoy to be built without our wrapper, which would clearly be not what we want

ci/do_ci.sh
```diff
@@ -4,29 +4,26 @@ bazel fetch //source/exe:envoy-static
 
 SOURCE_DIR="$(bazel info workspace)"
 
+
+ENVOY_DEPENDENCY=$(awk '/envoy = dict/{p=1}; /commit/{print $3; exit}' bazel/repository_locations.bzl | sed -E 's|"([a-z0-9]+)",|\1|g')
```
jbohanon, line 8: would probably be good to comment what this awk/sed blackmagic is doing

source/exe/BUILD
```diff
@@ -10,9 +10,16 @@ load(
...
+alias(
+    name = "envoy",
+    actual = ":envoy-static",
+    visibility = ["//visibility:public"],
+)
+
 envoy_cc_binary(
     name = "envoy-static",
     repository = "@envoy",
+    visibility = ["//visibility:public"],
```
nfuden, line 22: why does this become public visibility?

distribution/binary/BUILD
```diff
@@ -0,0 +1,42 @@
...
+
+licenses(["notice"])  # Apache 2
+
+envoy_package()
+
+bundled(
+    name = "envoy",
+    mode = "opt",
+    visibility = ["//visibility:public"],
+    stripopts = ["--strip-all"],
```
nfuden, line 14: yeah this is why I am unsure if the dbg will work but per our chat if we put this in 1.27 then it wont block edge

source/exe/BUILD
jbohanon: I was throwing spaghetti with commits aba6fc6, afdf822, and 57d1e14. The latter is the one that unblocked so the former two can probably be reverted

ci/do_ci.sh
```diff
@@ -51,24 +36,28 @@ export BAZEL_EXTRA_TEST_OPTIONS="--test_env=ENVOY_IP_TEST_VERSIONS=v4only --test
...
+sed -i 's|"//contrib/..."||' "$UPSTREAM_ENVOY_SRCDIR/ci/do_ci.sh"
+
+bash -x "$UPSTREAM_ENVOY_SRCDIR/ci/do_ci.sh" "$@"
+
+ENVOY_GLOO_BIN_DIR='linux/amd64/build_envoy_release'
+ENVOY_GLOO_STRIPPED_DIR="${ENVOY_GLOO_BIN_DIR}_stripped"
+mkdir -p "$ENVOY_GLOO_BIN_DIR"
+mkdir -p "$ENVOY_GLOO_STRIPPED_DIR"
+
+bazel run @envoy//tools/zstd:zstd -- --stdout -d /build/envoy/x64/bin/release.tar.zst \
```
nfuden, line 56: ugh I feel silly. I like this easy way to get zstd working like upstream has!

PR #272, opened 2023-09-28
ci: Update the intermediate image that gets emitted to no longer rely on alpine due to glibc issues

nfuden:
Move to ubuntu for now.
This can be migrated to distroless in solo-projects if need be in the future but would need to have ubuntu for symbols anyways if we want to follow upstream's behavior.

https://gitlab.alpinelinux.org/alpine/tsc/-/issues/43#note_306270
https://github.com/sgerrand/alpine-pkg-glibc/issues/207#issuecomment-1707209887
https://github.com/sgerrand/alpine-pkg-glibc/issues/176

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,7 +1,10 @@
-# This file was inspired by mesh Dockerfile:
-# https://github.com/solo-io/gloo-mesh-enterprise/blob/c40b21c3ef260de1acc92ba25bd1794b17074c8c/docker/Dockerfile.ubuntu#L4
-# as it previously was based off of envoys old alpine setup. 
-FROM ubuntu:focal-20220826
+# This file was inspired by a combination of the following:
+# Istio and platform's mesh derivative approach
+# Uses upstream envoy's ubuntu versioning
+# Consuemd by edge to buld gateway-proxy and gloo pods
+# May be used as is or as a builder intermediate (e.g. for distroless)
+# ON_MINOR_UPDATE: Pull from https://github.com/envoyproxy/envoy/blob/4d46da0bba54dfb849d8bf68b600e53d87310a1a/ci/Dockerfile-envoy#L1-L2
+FROM ubuntu:focal
 
 ENV loglevel=info
 
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,17 +1,23 @@
-# This file was inspired by envoy Dockerfile:
-# https://github.com/envoyproxy/envoy/blob/445a67344ffda0c8828c8e438e463fcaa7878434/ci/Dockerfile-envoy-alpine
-
-FROM frolvlad/alpine-glibc:alpine-3.17_glibc-2.34
+# This file was inspired by mesh Dockerfile:
+# https://github.com/solo-io/gloo-mesh-enterprise/blob/c40b21c3ef260de1acc92ba25bd1794b17074c8c/docker/Dockerfile.ubuntu#L4
+# as it previously was based off of envoys old alpine setup. 
+FROM ubuntu:focal-20220826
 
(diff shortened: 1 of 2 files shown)


ci/Dockerfile
```diff
@@ -1,17 +1,23 @@
-# This file was inspired by envoy Dockerfile:
-# https://github.com/envoyproxy/envoy/blob/445a67344ffda0c8828c8e438e463fcaa7878434/ci/Dockerfile-envoy-alpine
-
-FROM frolvlad/alpine-glibc:alpine-3.17_glibc-2.34
+# This file was inspired by mesh Dockerfile:
+# https://github.com/solo-io/gloo-mesh-enterprise/blob/c40b21c3ef260de1acc92ba25bd1794b17074c8c/docker/Dockerfile.ubuntu#L4
+# as it previously was based off of envoys old alpine setup. 
+FROM ubuntu:focal-20220826
```
ben-taussig-solo, line 4: Thanks for including the link to the mesh Dockerfile above. Based on the Ubuntu Docker hub page, it looks like `focal-20230801` is the latest 20.04 image release -- do you think it makes sense to bump to that, or is it more important to use the same version as platform? https://hub.docker.com/_/ubuntu

ci/Dockerfile
```diff
@@ -1,17 +1,23 @@
...
 
 ENV loglevel=info
 
-RUN apk upgrade --update-cache \
-    && apk add dumb-init ca-certificates \
-    && rm -rf /var/cache/apk/*
+ENV DEBIAN_FRONTEND=noninteractive
+
+# hadolint ignore=DL3005,DL3008
```
ben-taussig-solo, line 10: nit: Are we currently using hadolint in envoy-gloo?

ben-taussig-solo:
Looks good, though it would be nice to see confirmation of this being consumed + passing CI in Gloo OSS before merging

ci/Dockerfile
nfuden: Mimicks istio

nfuden:
yep scanned a built version of edge and the only issues that were of critical were from the non-in docker pieces that were due to my go version (which I then updated). Scan the images from  https://github.com/solo-io/gloo/pull/8730 to see this

ci/Dockerfile
```diff
@@ -1,17 +1,26 @@
...
+# May be used as is or as a builder intermediate (e.g. for distroless)
+# ON_MINOR_UPDATE: Pull from https://github.com/envoyproxy/envoy/blob/4d46da0bba54dfb849d8bf68b600e53d87310a1a/ci/Dockerfile-envoy#L1-L2
+FROM ubuntu:focal
 
 ENV loglevel=info
 
-RUN apk upgrade --update-cache \
-    && apk add dumb-init ca-certificates \
-    && rm -rf /var/cache/apk/*
+ENV DEBIAN_FRONTEND=noninteractive
```
ben-taussig-solo, line 11: I missed this on the first pass -- can we get away with removing the various `-y`s in the first `RUN` command if we have this set?

nfuden: Mimicked from mesh. We may be able to but I was afraid of chestertons fence.

nfuden: Resolving per your research shared in slack: https://unix.stackexchange.com/a/701288
https://solo-io-corp.slack.com/archives/C011H3TU17T/p1696270549377469?thread_ts=1696259513.309839&cid=C011H3TU17T

ben-taussig-solo: I don't know how this interacts with apt (see stackoverflow post here: https://unix.stackexchange.com/a/701288) so I say we leave it

ashishb-solo (approved):

PR #273, opened 2023-09-28
ci: Update the intermediate image that gets emitted to no longer rely…

nfuden:
Move to ubuntu for now.
This can be migrated to distroless in solo-projects if need be in the future but would need to have ubuntu for symbols anyways if we want to follow upstream's behavior.

https://gitlab.alpinelinux.org/alpine/tsc/-/issues/43#note_306270
https://github.com/sgerrand/alpine-pkg-glibc/issues/207#issuecomment-1707209887
https://github.com/sgerrand/alpine-pkg-glibc/issues/176

### DIFF
--- a/changelog/v1.27.0-patch2/updatebuild-image.yaml
+++ b/changelog/v1.27.0-patch2/updatebuild-image.yaml
@@ -1,0 +1,18 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/solo-projects/issues/5344
+    resolvesIssue: false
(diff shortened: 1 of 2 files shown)


ben-taussig-solo:
I don't have anything to add here apart from the commentary on the PR targeting `main` (https://github.com/solo-io/envoy-gloo-ee/pull/650)

One point -- could you clarify why both `release-v1.26` and `main` seem to rely on envoy v1.26?

ashishb-solo (approved):

PR #274, opened 2023-10-03
add unified matcher support to transformation filter

yuval-k:


source/extensions/filters/http/transformation/transformation_filter_config.h
```diff
@@ -50,21 +43,26 @@ class TransformationFilterConfig : public FilterConfig {
...
   std::string name() const override {
     return SoloHttpFilterNames::get().Transformation;
   }
 
   bool logRequestResponseInfo() const {
     return log_request_response_info_;
   }
+protected:
+
+  const std::vector<MatcherTransformerPair> &transformerPairs() const override {
```
nfuden, line 55: why does this need to be protected?

yuval-k: it was public before and it confused me, so I moved to protected so its more clear it is not meant to be used by non sub-classes

nfuden (approved):

PR #289, opened 2023-10-26
Tap filter v1.26.x

ashishb-solo:


nfuden:
is this still needed?

ashishb-solo:
> is this still needed?

i'm pretty sure it isn't

PR #301, opened 2024-01-25
[WIP] Add regex replace functionality to Transformation Filter Extractors

ben-taussig-solo:
# Description
 - WIP
 - I need to evaluate how this functionality interacts with:
   - `mergeExtractorsToBody`
   - the `extraction` inja callback (which lets a user reference an extraction by name)
   before this PR is ready for review
 - However, I thought I'd put this up some interested parties can take a look at where this work is at currently

### DIFF
--- a/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
+++ b/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
@@ -171,6 +171,12 @@ message Extraction {
   // If your regex contains capturing groups, use this field to determine which
   // group should be selected.
   uint32 subgroup = 3;
+
+  // The string to replace the matched portion of the source with
+  google.protobuf.StringValue replacement_text = 5;
+
+  // If set to true, all matches of the regex in the source will be replaced by the replacement_text.
+  bool replace_all = 6;
 }
 
 // Defines a transformation template.
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -56,7 +56,10 @@ getHeader(const Http::RequestOrResponseHeaderMap &header_map,
 Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
     : headername_(extractor.header()), body_(extractor.has_body()),
       group_(extractor.subgroup()),
-      extract_regex_(Solo::Regex::Utility::parseStdRegex(extractor.regex())) {
+      extract_regex_(Solo::Regex::Utility::parseStdRegex(extractor.regex())),
+      has_replacement_text_(extractor.has_replacement_text()),
+      replacement_text_(extractor.replacement_text().value()),
+      replace_all_(extractor.replace_all()) {
   // mark count == number of sub groups, and we need to add one for match number
   // 0 so we test for < instead of <= see:
   // http://www.cplusplus.com/reference/regex/basic_regex/mark_count/
@@ -64,6 +67,26 @@ Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
     throw EnvoyException(
         fmt::format("group {} requested for regex with only {} sub groups",
                     group_, extract_regex_.mark_count()));
+  }
+
+  // if replace_all is set, we must have replacement text
+  if (replace_all_ && !has_replacement_text_) {
+    throw EnvoyException(
+        fmt::format("replace_all set but no replacement text provided"));
+  }
+
+  // if replace_all is set, subgroup should be 0
+  if (replace_all_ && group_ != 0) {
+    throw EnvoyException(
+        fmt::format("replace_all set but subgroup is not 0"));
+  }
+
+  // extractionFunc is either replaceValue or extractValue depending on whether 
+  // replacement_text_ is empty or not
+  if (has_replacement_text_) {
+    extraction_func_ = std::bind(&Extractor::replaceValue, this, _1, _2);
+  } else {
+    extraction_func_ = std::bind(&Extractor::extractValue, this, _1, _2);
   }
 }
 
@@ -74,13 +97,14 @@ Extractor::extract(Http::StreamFilterCallbacks &callbacks,
   if (body_) {
     const std::string &string_body = body();
     absl::string_view sv(string_body);
-    return extractValue(callbacks, sv);
+    return extraction_func_(callbacks, sv);
   } else {
     const Http::HeaderMap::GetResult header_entries = getHeader(header_map, headername_);
     if (header_entries.empty()) {
       return "";
     }
-    return extractValue(callbacks, header_entries[0]->value().getStringView());
+    const auto &header_value = header_entries[0]->value().getStringView();
+    return extraction_func_(callbacks, header_value);
   }
 }
 
@@ -103,6 +127,86 @@ Extractor::extractValue(Http::StreamFilterCallbacks &callbacks,
     ENVOY_STREAM_LOG(debug, "extractor regex did not match input", callbacks);
   }
   return "";
+}
+
+// Match a regex against the input value and replace the matched subgroup with the replacement_text_ value
+// writes the result to replaced_value_ and returns a absl::string_view to it
+// if replace_all_ is true, __all__ substrings matching the regex in the input value will be replaced
+// with the replacement_text_ value
+absl::string_view
+Extractor::replaceValue(Http::StreamFilterCallbacks &callbacks,
+                        absl::string_view value) const {
+  if (replace_all_) {
+    return replaceAllValues(callbacks, value);
+  } else {
+    return replaceIndividualValue(callbacks, value);
+  }
+}
+
+absl::string_view
+Extractor::replaceIndividualValue(Http::StreamFilterCallbacks &callbacks,
+                                  absl::string_view value) const {
+  std::match_results<absl::string_view::const_iterator> regex_result;
+
+  // if there are no matches, return an empty string
+  if (!std::regex_search(value.begin(), value.end(), regex_result, extract_regex_)) {
+    ENVOY_STREAM_LOG(debug, "replaceValue: extractor regex did not match input", callbacks);
+    return ""
+  }
+
+  // if the subgroup specified is greater than the number of subgroups in the regex, return an empty string
+  if (group_ >= regex_result.size()) {
+    ASSERT("no such group in the regex");
+    ENVOY_STREAM_LOG(debug, "replaceValue: invalid group specified for regex", callbacks);
+    return "";
+  }
+
+  // if the regex doesn't match the entire input value, return an empty string
+  if (regex_result[0].length() != long(value.length())) {
+    ENVOY_STREAM_LOG(debug, "replaceValue: Regex did not match entire input value. This is not allowed when replace_all is false", callbacks);
+    return "";
+  }
+
+  // Create a new string with the input value
+  std::string replaced(value.begin(), value.end());
+
+  // Replace the specified subgroup with the replacement_text_ value
+  const auto &sub_match = regex_result[group_];
+  auto start_pos = sub_match.first - value.begin();
+  auto length = sub_match.length();
+  replaced.replace(start_pos, length, replacement_text_);
+
+  // Store the replaced string in replaced_value_ and return an absl::string_view to it
+  replaced_value_ = std::move(replaced);
+  return absl::string_view(replaced_value_);
+}
+
+absl::string_view
+Extractor::replaceAllValues(Http::StreamFilterCallbacks &callbacks,
+                            absl::string_view value) const {
+  std::string input(value.begin(), value.end());
+  std::string replaced;
+  bool matchFound = false;
+
+  // create an iterator to search for matches of extract_regex_ in the input string
+  std::sregex_iterator it(input.begin(), input.end(), extract_regex_);
+  std::sregex_iterator end_it; // default end iterator for comparison
+
+  // set matchFound to true if at least one match is found
+  if (it != end_it) {
+      matchFound = true; 
+  }
+
+  if (!matchFound) {
+    return "";
+  }
+
+  // If a match was found, replace all instances of the regex in the input value with the replacement_text_ value
+  replaced = std::regex_replace(input, extract_regex_, replacement_text_);
+
+  // Store the replaced string in replaced_value_ and return an absl::string_view to it
+  replaced_value_ = std::move(replaced);
+  return absl::string_view(replaced_value_);
 }
 
 // A TransformerInstance is constructed by the InjaTransformer constructor at config time
--- a/source/extensions/filters/http/transformation/inja_transformer.h
+++ b/source/extensions/filters/http/transformation/inja_transformer.h
@@ -25,6 +25,7 @@ namespace HttpFilters {
 namespace Transformation {
 
 using GetBodyFunc = std::function<const std::string &()>;
+using ExtractionFunc = std::function<absl::string_view(Http::StreamFilterCallbacks &callbacks, absl::string_view value)>;
 
 struct ThreadLocalTransformerContext : public ThreadLocal::ThreadLocalObject {
 public:
@@ -86,11 +87,23 @@ public:
 private:
   absl::string_view extractValue(Http::StreamFilterCallbacks &callbacks,
                                  absl::string_view value) const;
+  absl::string_view replaceValue(Http::StreamFilterCallbacks &callbacks,
+                                 absl::string_view value) const;
+  absl::string_view replaceIndividualValue(Http::StreamFilterCallbacks &callbacks,
+                                           absl::string_view value) const;
+  absl::string_view replaceAllValues(Http::StreamFilterCallbacks &callbacks,
+                                     absl::string_view value) const;
 
   const Http::LowerCaseString headername_;
   const bool body_;
   const unsigned int group_;
   const std::regex extract_regex_;
+  const bool has_replacement_text_;
+  const std::string replacement_text_;
+  const bool replace_all_;
+
+  ExtractionFunc extraction_func_;
+  mutable std::string replaced_value_;
 };
 
 class InjaTransformer : public Transformer {
--- a/test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
+++ b/test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
@@ -1,0 +1,442 @@
+#include "source/extensions/filters/http/solo_well_known_names.h"
+#include "source/extensions/filters/http/transformation/inja_transformer.h"
+#include "source/common/common/base64.h"
+#include "source/common/common/random_generator.h"
+
+#include "test/mocks/common.h"
+#include "test/mocks/http/mocks.h"
+#include "test/mocks/server/mocks.h"
+#include "test/mocks/thread_local/mocks.h"
+#include "test/mocks/upstream/mocks.h"
+#include "test/test_common/environment.h"
+
+#include "fmt/format.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+#include <memory>
+
+using testing::_;
+using testing::AtLeast;
+using testing::HasSubstr;
+using testing::Invoke;
+using testing::Return;
+using testing::ReturnPointee;
+using testing::ReturnRef;
+using testing::SaveArg;
+using testing::WithArg;
+
+using json = nlohmann::json;
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Transformation {
+
+using TransformationTemplate =
+    envoy::api::v2::filter::http::TransformationTemplate;
+
+namespace {
+std::function<const std::string &()> empty_body = [] { return EMPTY_STRING; };
+}
+
+class TransformerInstanceTest : public testing::Test {
+protected:
+  NiceMock<Random::MockRandomGenerator> rng_;
+  NiceMock<ThreadLocal::MockInstance> tls_;
+};
+
+void fill_slot(
+      ThreadLocal::SlotPtr& slot,
+      const Http::RequestOrResponseHeaderMap &header_map,
+      const Http::RequestHeaderMap *request_headers,
+      GetBodyFunc &body,
+      const std::unordered_map<std::string, absl::string_view> &extractions,
+      const nlohmann::json &context,
+      const std::unordered_map<std::string, std::string> &environ,
+      const envoy::config::core::v3::Metadata *cluster_metadata) {
+  slot->set([](Event::Dispatcher&) -> ThreadLocal::ThreadLocalObjectSharedPtr {
+          return std::make_shared<ThreadLocalTransformerContext>();
+  });
+  auto& typed_slot = slot->getTyped<ThreadLocalTransformerContext>();
+  typed_slot.header_map_ = &header_map;
+  typed_slot.request_headers_ = request_headers;
+  typed_slot.body_ = &body;
+  typed_slot.extractions_ = &extractions;
+  typed_slot.context_ = &context;
+  typed_slot.environ_ = &environ;
+  typed_slot.cluster_metadata_ = cluster_metadata;
+}
+
+TEST(Extraction, ExtractAndReplaceValueFromBodySubgroup) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  extractor.set_regex(".*(body)");
+  extractor.set_subgroup(1);
+  auto replacement_text = "BAZ";
+  extractor.mutable_replacement_text()->set_value(replacement_text);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body("not json body");
+  GetBodyFunc bodyfunc = [&body]() -> const std::string & { return body; };
+  std::string res(Extractor(extractor).extract(callbacks, headers, bodyfunc));
+
+  EXPECT_EQ("not json BAZ", res);
+}
+
+TEST(Extraction, ExtractAndReplaceValueFromFullBody) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  extractor.set_regex(".*");
+  extractor.set_subgroup(0);
+  auto replacement_text = "BAZ";
+  extractor.mutable_replacement_text()->set_value(replacement_text);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body("not json body");
+  GetBodyFunc bodyfunc = [&body]() -> const std::string & { return body; };
+  std::string res(Extractor(extractor).extract(callbacks, headers, bodyfunc));
+
+  EXPECT_EQ("BAZ", res);
+}
+
+// TODO: EDGE CASE !!!
+TEST(Extraction, ExtractAndReplaceAllFromFullBody) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  extractor.set_regex(".*");
+  extractor.set_subgroup(0);
+  extractor.set_replace_all(true);
+  auto replacement_text = "BAZ";
+  extractor.mutable_replacement_text()->set_value(replacement_text);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body("not json body");
+  GetBodyFunc bodyfunc = [&body]() -> const std::string & { return body; };
+  std::string res(Extractor(extractor).extract(callbacks, headers, bodyfunc));
+
+  // Note to self/reviewers: this is the current behavior, which is a kind of 
+  // confusing edge case in std::regex_replace when the regex is .*
+  // apparently, this regex matches the whole input string __AND__ the 
+  // line ending, so the replacement is applied twice
+  EXPECT_EQ("BAZBAZ", res);
+}
+
+TEST(Extraction, AttemptReplaceFromPartialMatch) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  // Unless replace_all is set, we require regexes to match the entire target string
+  // because this only matches a substring, it should not be replaced
+  extractor.set_regex("body");
+  extractor.set_subgroup(0);
+  auto replacement_text = "BAZ";
+	extractor.mutable_replacement_text()->set_value(replacement_text);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body("not json body");
+  GetBodyFunc bodyfunc = [&body]() -> const std::string & { return body; };
+  std::string res(Extractor(extractor).extract(callbacks, headers, bodyfunc));
+
+  EXPECT_EQ("", res);
+}
+
+TEST(Extraction, AttemptReplaceFromPartialMatchNonNilSubgroup) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  // Unless replace_all is set, we require regexes to match the entire target string
+  // because this only matches a substring, it should not be replaced
+  // Note -- the subgroup in the regex is introduced here so that this config is not
+  // rejected when constructing the extractor
+  extractor.set_regex("(body)");
+  extractor.set_subgroup(1);
+  auto replacement_text = "BAZ";
+	extractor.mutable_replacement_text()->set_value(replacement_text);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body("not json body");
+  GetBodyFunc bodyfunc = [&body]() -> const std::string & { return body; };
+  std::string res(Extractor(extractor).extract(callbacks, headers, bodyfunc));
+
+  EXPECT_EQ("", res);
+}
+
+
+TEST(Extraction, ReplaceFromFullLiteralMatch) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  // We require regexes to match the entire target string
+  // because this only matches a substring, it should not be replaced
+  extractor.set_regex("not json body");
+  extractor.set_subgroup(0);
+  auto replacement_text = "BAZ";
+	extractor.mutable_replacement_text()->set_value(replacement_text);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body("not json body");
+  GetBodyFunc bodyfunc = [&body]() -> const std::string & { return body; };
+  std::string res(Extractor(extractor).extract(callbacks, headers, bodyfunc));
+
+  EXPECT_EQ("BAZ", res);
+}
+
+TEST(Extraction, AttemptToReplaceFromInvalidSubgroup) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  extractor.set_regex(".*");
+  extractor.set_subgroup(1);
+  auto replacement_text = "BAZ";
+	extractor.mutable_replacement_text()->set_value(replacement_text);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body("not json body");
+  GetBodyFunc bodyfunc = [&body]() -> const std::string & { return body; };
+  EXPECT_THROW_WITH_MESSAGE(Extractor(extractor).extract(callbacks, headers, bodyfunc), EnvoyException, "group 1 requested for regex with only 0 sub groups");
+}
+
+TEST(Extraction, NestedSubgroups) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  extractor.set_regex(".*(not (json) body)");
+  extractor.set_subgroup(2);
+  auto replacement_text = "BAZ";
+	extractor.mutable_replacement_text()->set_value(replacement_text);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body("not json body");
+  GetBodyFunc bodyfunc = [&body]() -> const std::string & { return body; };
+  std::string res(Extractor(extractor).extract(callbacks, headers, bodyfunc));
+
+  EXPECT_EQ("not BAZ body", res);
+}
+
+TEST(Extraction, SubgroupUnset) {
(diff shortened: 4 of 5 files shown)


source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -105,6 +129,86 @@ Extractor::extractValue(Http::StreamFilterCallbacks &callbacks,
...
+  // create an iterator to search for matches of extract_regex_ in the input string
+  std::sregex_iterator it(input.begin(), input.end(), extract_regex_);
+  std::sregex_iterator end_it; // default end iterator for comparison
+
+  // set matchFound to true if at least one match is found
+  if (it != end_it) {
+      matchFound = true; 
+  }
+
+  if (!matchFound) {
```
ben-taussig-solo, line 200: TODO: add logging here

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
```diff
@@ -0,0 +1,442 @@
...
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  std::string body("not json body");
+  GetBodyFunc bodyfunc = [&body]() -> const std::string & { return body; };
+  std::string res(Extractor(extractor).extract(callbacks, headers, bodyfunc));
+
+  // Note to self/reviewers: this is the current behavior, which is a kind of 
+  // confusing edge case in std::regex_replace when the regex is .*
+  // apparently, this regex matches the whole input string __AND__ the 
+  // line ending, so the replacement is applied twice
+  EXPECT_EQ("BAZBAZ", res);
```
ben-taussig-solo, line 127: Note to reviewers: this edge case frustrates me, trying to figure out how to handle it at the moment

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -86,11 +87,23 @@ class Extractor : Logger::Loggable<Logger::Id::filter> {
...
+                                           absl::string_view value) const;
+  absl::string_view replaceAllValues(Http::StreamFilterCallbacks &callbacks,
+                                     absl::string_view value) const;
 
   const Http::LowerCaseString headername_;
   const bool body_;
   const unsigned int group_;
   const std::regex extract_regex_;
+  const bool has_replacement_text_;
+  const std::string replacement_text_;
```
ben-taussig-solo, line 102: `has_replacement_text_` and `replacement_text_` are each derived from the `google.protobuf.stringValue` API field  `replacement_text`. We could have a single `google::protobuf::stringValue` field on the extractor that is 1:1 with the API field, but I think having a separate bool/string is cleaner. That being said, I'm open to changing this if reviewers have strong feelings

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -65,6 +68,26 @@ Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
...
+  }
+
+  // if replace_all is set, subgroup should be 0
+  if (replace_all_ && group_ != 0) {
+    throw EnvoyException(
+        fmt::format("replace_all set but subgroup is not 0"));
+  }
+
+  // extractionFunc is either replaceValue or extractValue depending on whether 
+  // replacement_text_ is empty or not
```
ben-taussig-solo, line 85: Nit: 
```suggestion
  // we have replacement_text_
```

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: resolved here: 3d57195489d9de1f2f7c29b059b3746053d97bca

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -171,6 +171,12 @@ message Extraction {
   // If your regex contains capturing groups, use this field to determine which
   // group should be selected.
   uint32 subgroup = 3;
+
+  // The string to replace the matched portion of the source with
+  google.protobuf.StringValue replacement_text = 5;
+
+  // If set to true, all matches of the regex in the source will be replaced by the replacement_text.
```
ben-taussig-solo, line 178: TODO: update commentary to indicate that this is mutually exclusive with `subgroup`

ashishb-solo, line 178: Could you explain what it does if it's false? Does it only replace the first instance then?

ben-taussig-solo: That is essentially correct. We call this out in a couple of places in the docs, but by default an extractor's `regex` field needs to match the entire input (if you scroll down a bit [here](https://docs.solo.io/gloo-edge/1.7.23/guides/traffic_management/request_processing/transformations/#extractors) there's a comment speling this out)

```yaml
    # The regex must match the ENTIRE body, else the extraction will not work
    # The [\s\S]* allows us to match any character, making the regex match the entire source.
    regex: '[\s\S]*id: ([0-9]{4})[\s\S]*'
```

**so if `replace_all` is false and `replacement_text` is set, we replace the selected subgroup with the `replacement_text`** (this is the case you're calling out)

if `replace_all` is false and `replacement_text` is not set, we just use the pre-existing extractor functionality (i.e., extract the value of the selected subgroup)

if `replace_all` is true, we relax the restriction that the regex needs to match the entire source (I have to say, I don't love that restrictions around the regex change when this field is set). In this case, the regex should only partially match the input value. All matches against the input value will have their content replaced by `replacement_text`

note that if `replace_all` is true, `replacement_text` must be set 
 - I am going to validate that I test against this

ben-taussig-solo: I was worried about it breaking existing extractor configuration but I haven't actually tried it out -- let me take a look and I'll report back

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -153,6 +153,19 @@ message Transformation {
 // Extractions can be used to extract information from the request/response.
 // The extracted information can then be referenced in template fields.
 message Extraction {
+  // Define the Mode enum within the Extraction message
```
ben-taussig-solo, line 156: Nit: probably unnecessary

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
ben-taussig-solo: I updated the API to include a mode selection which should hopefully clarify some of these issues here: https://github.com/solo-io/envoy-gloo/pull/301/commits/3554ba1e82997a59633c4a17a7bc4011f1ac4ac0

```yaml
  enum Mode {
    // Default mode. Extract the value of the subgroup-th capturing group.
    EXTRACT = 0; 
    // Replace the value of the subgroup-th capturing group with the replacement_text.
    // Note: replacement_text must be set for this mode.
    SINGLE_REPLACE = 1;
    // Replace all matches of the regex in the source with the replacement_text.
    // Note: replacement_text must be set for this mode.
    // Note: subgroup is ignored for this mode. configuration will fail if subgroup is set.
    // Note: restrictions on the regex are different for this mode. See
    REPLACE_ALL = 2;
  }
```

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -153,6 +153,19 @@ message Transformation {
...
+  enum Mode {
+    // Default mode. Extract the value of the subgroup-th capturing group.
+    EXTRACT = 0; 
+    // Replace the value of the subgroup-th capturing group with the replacement_text.
+    // Note: replacement_text must be set for this mode.
+    SINGLE_REPLACE = 1;
+    // Replace all matches of the regex in the source with the replacement_text.
+    // Note: replacement_text must be set for this mode.
+    // Note: subgroup is ignored for this mode. configuration will fail if subgroup is set.
+    // Note: restrictions on the regex are different for this mode. See
```
ben-taussig-solo, line 166: Complete this comment

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: This comment no longer exists, closing the thread

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
ben-taussig-solo: Updated here: f747a124c15c56a09d6436607f4efc992e37859d

```diff
-  // Define the Mode enum within the Extraction message
+  // The mode of operation for the extraction.
```

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
ben-taussig-solo: Updated here: https://github.com/solo-io/envoy-gloo/commit/f747a124c15c56a09d6436607f4efc992e37859d

```diff
-    // Note: restrictions on the regex are different for this mode. See
+    // Note: restrictions on the regex are different for this mode. See the regex field for more details.
```

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -105,6 +132,77 @@ Extractor::extractValue(Http::StreamFilterCallbacks &callbacks,
...
+
+  // if there are no matches, return an empty string
+  if (!std::regex_search(value.begin(), value.end(), regex_result, extract_regex_)) {
+    ENVOY_STREAM_LOG(debug, "replaceValue: extractor regex did not match input", callbacks);
+    return "";
+  }
+
+  // if the subgroup specified is greater than the number of subgroups in the regex, return an empty string
+  if (group_ >= regex_result.size()) {
+    ASSERT("no such group in the regex");
```
jbohanon, line 150: curious, what does `ASSERT`ing a string literal do here?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -105,6 +132,77 @@ Extractor::extractValue(Http::StreamFilterCallbacks &callbacks,
...
+    matchFound = true; 
+  }
+
+  if (!matchFound) {
+    ENVOY_STREAM_LOG(debug, "replaceAllValues: extractor regex did not match input", callbacks);
+    return "";
+  }
+
+  // If a match was found, replace all instances of the regex in the input value with the replacement_text_ value
+  replaced = std::regex_replace(input, extract_regex_, replacement_text_);
```
ashishb-solo, line 199: similar performance concerns here since we have had issues with `std::regex` in the past

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -83,14 +84,31 @@ class Extractor : Logger::Loggable<Logger::Id::filter> {
                             const Http::RequestOrResponseHeaderMap &header_map,
                             GetBodyFunc &body) const;
 
+  // Matching enum with the protobuf definition
+  enum class Mode {
+    EXTRACT = 0,
+    SINGLE_REPLACE = 1,
+    REPLACE_ALL = 2
```
jbohanon, line 91: Could we use the proto enum instead? It's one less thing to keep track of if the proto enum grows in the future

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: I think we'll segfault if we hit this... I don't think we should ever hit this (maybe in the no match case 😬) but I will confirm and do something else if that is in fact the case

source/extensions/filters/http/transformation/inja_transformer.h
ben-taussig-solo: definitely, good call

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -105,6 +132,77 @@ Extractor::extractValue(Http::StreamFilterCallbacks &callbacks,
...
+  // if the subgroup specified is greater than the number of subgroups in the regex, return an empty string
+  if (group_ >= regex_result.size()) {
+    ASSERT("no such group in the regex");
+    ENVOY_STREAM_LOG(debug, "replaceValue: invalid group specified for regex", callbacks);
+    return "";
+  }
+
+  // if the regex doesn't match the entire input value, return an empty string
+  if (regex_result[0].length() != long(value.length())) {
+    ENVOY_STREAM_LOG(debug, "replaceValue: Regex did not match entire input value. This is not allowed when replace_all is false", callbacks);
```
ben-taussig-solo, line 157: TODO: update comments with reference to no-longer-existing `replace_all` bool

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -105,6 +132,77 @@ Extractor::extractValue(Http::StreamFilterCallbacks &callbacks,
...
+
+  // Replace the specified subgroup with the replacement_text_ value
+  const auto &sub_match = regex_result[group_];
+  auto start_pos = sub_match.first - value.begin();
+  auto length = sub_match.length();
+  replaced.replace(start_pos, length, replacement_text_);
+
+  // Store the replaced string in replaced_value_ and return an absl::string_view to it
+  replaced_value_ = std::move(replaced);
+  return absl::string_view(replaced_value_);
```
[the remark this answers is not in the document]
ben-taussig-solo: > What would happen if the replacement text is longer than the regex
  - This is a very good point. I will add a test against this. I don't know how `.replace` will behave + I do expect we would need an additional allocation even if the results are correct. I'm going to think about coming up w/ a more robust way to handle this

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -540,6 +540,42 @@ TEST_F(TransformerTest, transformMergeExtractorsToBody) {
   EXPECT_EQ("{\"ext1\":\"123\"}", res);
 }
 
+TEST_F(TransformerTest, transformMergeReplaceExtractorsToBody) {
```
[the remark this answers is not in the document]
ben-taussig-solo: There are some unit tests that do multiple replace, however there aren't any in `inja_transformer_test.cc` at the moment -- I'm happy to add some there if you think it's relevant

For example: https://github.com/solo-io/envoy-gloo/blob/f747a124c15c56a09d6436607f4efc992e37859d/test/extensions/filters/http/transformation/inja_transformer_replace_test.cc#L368-L386

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -65,6 +68,26 @@ Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
...
                     group_, extract_regex_.mark_count()));
   }
+
+  // if replace_all is set, we must have replacement text
+  if (replace_all_ && !has_replacement_text_) {
+    throw EnvoyException(
+        fmt::format("replace_all set but no replacement text provided"));
+  }
+
+  // if replace_all is set, subgroup should be 0
```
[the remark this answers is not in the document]
ben-taussig-solo: This is documented with the API update that added the mode selection switch:
```proto
Note: subgroup is ignored for this mode. configuration will fail if subgroup is set.
```
3554ba1e82997a59633c4a17a7bc4011f1ac4ac0

As such, I'm resolving this thread

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Some notes/questions:
 - Extractor regexes are compiled at config time, which is good (or really, the alternative is unacceptable)
   - https://github.com/solo-io/envoy-gloo/blob/7da296e3c690b02aa0fe1bc879950cdf0ea40ee9/source/extensions/filters/http/transformation/inja_transformer.cc#L59
 - The replaceAll path (i.e. the one we're discussing here) is not maximally efficient -- it does an efficient check to identify whether there are any matches, and then does the call you've highlighted here, which will replace all matches.
   - I considered extending the efficient check to identify matches to identify all matches, after which we could manually replace matches with the replacement_text_, but I chose not to as it seemed like it would be difficult to implement/read/maintain, and would potentially be prone to bugs and edge cases. But I could go that route if the performance of `regex_replace` is problematically bad. That being said I think I would prefer to get some performance metrics before doing so
 - Can you link any issues or PRs you can recall where there's discussion about `std::regex` performance issues?

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
```diff
@@ -0,0 +1,442 @@
...
+
+TEST(Extraction, AttemptReplaceFromPartialMatchNonNilSubgroup) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  // Unless replace_all is set, we require regexes to match the entire target string
+  // because this only matches a substring, it should not be replaced
+  // Note -- the subgroup in the regex is introduced here so that this config is not
+  // rejected when constructing the extractor
```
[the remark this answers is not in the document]
ben-taussig-solo: I am going to reword this comment -- I had a hard time understanding what I was saying here until I remembered the example.

> What is the behavior when `subgroup` is not set? 
  - `subgroup` defaults to 0 when not explicitly set

The point here is that there's some config-time checking to make sure the configured `subgroup` doesn't exceed the number of capturing groups in a `regex`. e.g.:
 - `"body"` has no capturing groups, so we will reject config that attempts to set a `subgroup` higher than 0 (which refers to the entire input) if that is set as the `regex`
 - `"(body)"` has 1 capturing group, so we will reject config that attempts to set a `subgroup` higher than 1 if that is set as the `regex`. That being said, configuring a `subgroup` of 1 is legal

The regex here doesn't match the full input so we expect no match, and this test is trying to confirm that setting non-nil (i.e. non-default) subgroups doesn't affect that

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -56,7 +56,10 @@ getHeader(const Http::RequestOrResponseHeaderMap &header_map,
 Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
     : headername_(extractor.header()), body_(extractor.has_body()),
       group_(extractor.subgroup()),
-      extract_regex_(Solo::Regex::Utility::parseStdRegex(extractor.regex())) {
+      extract_regex_(Solo::Regex::Utility::parseStdRegex(extractor.regex())),
+      has_replacement_text_(extractor.has_replacement_text()),
+      replacement_text_(extractor.replacement_text().value()),
```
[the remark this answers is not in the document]
ben-taussig-solo: Thanks, I always have a hard time getting the semantics right with absl optional fields so this is much appreciated

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -65,6 +68,26 @@ Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
...
+  // if replace_all is set, we must have replacement text
+  if (replace_all_ && !has_replacement_text_) {
+    throw EnvoyException(
+        fmt::format("replace_all set but no replacement text provided"));
+  }
+
+  // if replace_all is set, subgroup should be 0
+  if (replace_all_ && group_ != 0) {
+    throw EnvoyException(
+        fmt::format("replace_all set but subgroup is not 0"));
```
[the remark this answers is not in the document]
ben-taussig-solo: Good catch, I should've highlighted this because I was unsure about this during implementation
 
Subgroup is really just not considered if we're in the `replace_all` mode. We definitely could just log a warning in that case.

The thing that influenced my decision to throw an exception is that the `regex` field has slightly different constraints depending on the mode selected (extract vs. single replace vs. replace all)

https://github.com/solo-io/envoy-gloo/blob/f747a124c15c56a09d6436607f4efc992e37859d/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto#L178-L187

My thinking is that I don't want users to feel like they can toggle between modes on the fly -- if you have a working single replace extraction, it will not work the way you expect if you switch the mode to replace all. In that case, if you specify a subgroup (which I expect is typically the case), the system will indicate that the configuration is invalid, hopefully leading the user to realize that they may need more significant changes to their configuration.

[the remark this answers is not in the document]
jbohanon: Aha, that makes a lot of sense. Thanks for the explanation. Given the anticipated use case of users hotswapping or copy/pasting config and changing the mode, I think it makes sense to enforce that. That said, I feel like that enforcement should occur on the control plane so that we are not throwing an exception on config. I see that we have precedent for doing that already, so I'm not super opinionated about it...

Def think we should enforce in the control plane though

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: resolved here: https://github.com/solo-io/envoy-gloo/pull/301/commits/db42db7f590726f6c8e2290d7b45645af23ee7a0

source/extensions/filters/http/transformation/inja_transformer.h
ben-taussig-solo: Resolved here: https://github.com/solo-io/envoy-gloo/pull/301/commits/66c09331678aecc7a3cadce6094d472edefd4ee0

source/extensions/filters/http/transformation/inja_transformer.h
ben-taussig-solo: Updated to use a std::optional here: https://github.com/solo-io/envoy-gloo/pull/301/commits/71000c6b9712d53722b6cb0cb4ad2549a2444f19

source/extensions/filters/http/transformation/inja_transformer.cc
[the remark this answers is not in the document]
ben-taussig-solo: Updated to use a std::optional here: https://github.com/solo-io/envoy-gloo/pull/301/commits/71000c6b9712d53722b6cb0cb4ad2549a2444f19

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Some notes:
 - I lifted this block from pre-existing extractor code, in `Extractor::extractValue`
https://github.com/solo-io/envoy-gloo/blob/f747a124c15c56a09d6436607f4efc992e37859d/source/extensions/filters/http/transformation/inja_transformer.cc#L121-L126
 - I have confirmed that the comment in the pre-existing code indicating that this block should never be entered. 
   - Specifically, the new test `AttemptReplaceFromPartialMatchNonNilSubgroup` validates this -- it confirms that `regex_result.size()` is equal to the number of capturing groups in the regex plus 1, even in the case where the regex does not match
https://github.com/solo-io/envoy-gloo/blob/f747a124c15c56a09d6436607f4efc992e37859d/test/extensions/filters/http/transformation/inja_transformer_replace_test.cc#L153-L174
 - I have added another test to validate this here: 0f64ffe89d845bd8f2242966033730165f3bdd50
 - I have updated the comment in the block you highlighted to match the comment from the pre-existing code, indicating that this block should never be reached as this condition is validated in the constructor: 1440a95b8a3ff6ad03c1d49c82b9d2f1e1732bb0

source/extensions/filters/http/transformation/inja_transformer.cc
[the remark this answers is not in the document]
ben-taussig-solo: Updated here to prevent strings from being reallocated: 55ddd1bfc4d9976bc953cab2dd0fcc874edd9405

```cpp
  // Create a new string with the maximum possible length after replacement
  auto max_possible_length = value.length() + replacement_text_.value().length();
  std::string replaced;
  replaced.reserve(max_possible_length);

  auto subgroup_start = regex_result[group_].first;
  auto subgroup_end = regex_result[group_].second;

  // Copy the initial part of the string until the match
  replaced.assign(value.begin(), subgroup_start);

  // Append the replacement text
  replaced += replacement_text_.value();

  // Append the remaining part of the string after the match
  replaced.append(subgroup_end, value.end());
```

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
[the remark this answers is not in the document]
ben-taussig-solo: I think the inline API documentation now appropriately clarifies regex/subgroup requirements for different modes of operation. Let me know if you think this is sufficient:

```proto
  // The regex field specifies the regular expression used for matching against the source content.
  // - In EXTRACT mode, the entire source must match the regex. The subgroup-th capturing group,
  //   if specified, determines which part of the match is extracted.
  // - In SINGLE_REPLACE mode, the regex also needs to match the entire source. The subgroup-th capturing group
  //   is targeted for replacement with the replacement_text.
  // - In REPLACE_ALL mode, the regex is applied repeatedly to find all occurrences within the source that match.
  //   Each matching occurrence is replaced with the replacement_text, and the subgroup field is not used.
  // This field is required, and if the regex does not match the source as per the selected mode, the result of
  // the extraction will be an empty value.
  string regex = 2;

  // If your regex contains capturing groups, use this field to determine which
  // group should be selected.
  // For EXTRACT and SINGLE_REPLACE, refers to the portion of the text
  // to extract/replace.
  // Config will be rejected if this is specified in REPLACE_ALL mode.
  uint32 subgroup = 3;
```

ben-taussig-solo:
Note to reviewers: I've put up a draft PR in Gloo OSS with changes to support this functionality: https://github.com/solo-io/gloo/pull/9124

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
```diff
@@ -0,0 +1,491 @@
...
+
+TEST(Extraction, NilReplace) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  extractor.set_regex(".*(body)");
+  extractor.set_subgroup(1);
+  auto replacement_text = "";
+	extractor.mutable_replacement_text()->set_value(replacement_text);
```
jbohanon, line 317: nit spacing?

jbohanon: multiple places in this file

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -637,6 +675,32 @@ TEST_F(InjaTransformerTest, DontParseBodyAndExtractFromIt) {
...
+  TransformationTemplate transformation;
+  transformation.set_parse_body_behavior(TransformationTemplate::DontParse);
+  transformation.set_advanced_templates(true);
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  extractor.set_regex("not ([\\-._[:alnum:]]+) body");
+  extractor.set_subgroup(1);
+  auto replacement_text = "JSON";
+  extractor.mutable_replacement_text()->set_value(replacement_text);
```
jbohanon, line 691: nit: this var can be inlined

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -637,6 +675,32 @@ TEST_F(InjaTransformerTest, DontParseBodyAndExtractFromIt) {
   EXPECT_EQ(body.toString(), "json");
 }
 
+TEST_F(InjaTransformerTest, DontParseBodyAndExtractFromReplacementText) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+  Buffer::OwnedImpl body("not json body");
+
+  TransformationTemplate transformation;
+  transformation.set_parse_body_behavior(TransformationTemplate::DontParse);
+  transformation.set_advanced_templates(true);
```
jbohanon, line 684: nit: I don't think this is necessary here? Not hurting anything, but may add confusion for future readers

jbohanon (commented):
Mostly nits; looking really good

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
```diff
@@ -173,6 +173,29 @@ TEST(Extraction, AttemptReplaceFromPartialMatchNonNilSubgroup) {
   EXPECT_EQ("", res);
 }
 
+TEST(Extraction, AttemptReplaceFromNoMatchNonNilSubgroup) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
```
ashishb-solo, line 174: nit: can replace this type definition with `ExtractionApi` as defined [here](https://github.com/solo-io/envoy-gloo/blob/55ddd1bfc4d9976bc953cab2dd0fcc874edd9405/source/extensions/filters/http/transformation/inja_transformer.h#L29)

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
ashishb-solo: does changing line 114 to `extractor.set_regex(".*$")` change anything? or maybe `"^.*$"`?

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
ben-taussig-solo: ugh this is a tabs vs. spaces thing, thanks for pointing this out

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -82,15 +84,23 @@ class Extractor : Logger::Loggable<Logger::Id::filter> {
...
 
   const Http::LowerCaseString headername_;
   const bool body_;
   const unsigned int group_;
   const std::regex extract_regex_;
+  const std::optional<std::string> replacement_text_;
+  const ExtractionApi::Mode mode_;
+
+  ExtractionFunc extraction_func_;
+  mutable std::string replaced_value_;
```
ashishb-solo, line 103: as far as i can see, `replaced_value_` is only ever written to, but never read. do correct me if i'm wrong, but if not, is it necessary to keep this private variable around?

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
```diff
@@ -0,0 +1,491 @@
...
+
+  EXPECT_EQ("", res);
+}
+
+TEST(Extraction, ReplaceFromFullLiteralMatch) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}};
+
+  envoy::api::v2::filter::http::Extraction extractor;
+  extractor.mutable_body();
+  // We require regexes to match the entire target string
```
ashishb-solo, line 194: this comment looks like it was copy-pasted - is it supposed to be here?

test/extensions/filters/http/transformation/inja_transformer_test.cc
ben-taussig-solo: We do this in most of the surrounding tests (e.g. if you expand the file and look at `DontParseBodyAndExtractFromIt`, we do the same thing there). This affects the way we reference the extraction in the transformation definition. 

Currently, the new test, similar to the surrounding tests, references the extraction as such:
```cpp
transformation.mutable_body()->set_text("{{extraction(\"param\")}}");
```

If we disabled advanced templates, we would have to instead do
```cpp
transformation.mutable_body()->set_text("{{ param }}");
```

I think it's worthwhile to leave this as-is to maintain consistency with the existing tests in this file

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
```diff
@@ -0,0 +1,491 @@
...
+
+  std::string res(Extractor(extractor).extract(callbacks, headers, bodyfunc));
+
+  EXPECT_EQ("bar bar bar", res);
+}
+
+TEST(Extraction, ReplaceAllNoMatch) {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"}, {":path", "/foo"}, {"foo", "bar"}};
+
+  // Note that the regex contains enough capture groups
```
ashishb-solo, line 452: does this comment belong here?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -105,6 +131,86 @@ Extractor::extractValue(Http::StreamFilterCallbacks &callbacks,
...
+
+  // set matchFound to true if at least one match is found
+  if (it != end_it) {
+    matchFound = true; 
+  }
+
+  if (!matchFound) {
+    ENVOY_STREAM_LOG(debug, "replaceAllValues: extractor regex did not match input", callbacks);
+    return "";
+  }
```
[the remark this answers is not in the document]
ben-taussig-solo: Thanks for identifying this, the method looks much cleaner now. Resolved here: 497739df9b1f47852629813439a485b9baf39907

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
ben-taussig-solo: Resolved here: e8f9e74fbd152723852eba90849fa6e58bed0daf

ashishb-solo (commented):
looks good - only one (potentially) major comment that i found (`replaced_value_` seems unnecessary), but looks good to me

test/extensions/filters/http/transformation/inja_transformer_test.cc
ben-taussig-solo: Resolved in the replace tests here: e8f9e74fbd152723852eba90849fa6e58bed0daf

Just realized this comment was on tests added in inja_transformer_test.cc -- I'll put up a commit to fix this soon

ben-taussig-solo: Resolved in inja_transformer_test.cc here: 0a7d48eefab514ce13debc4efd01c6fe0b4f32ce

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
ben-taussig-solo: Thanks, it was not supposed to be there. Resolved, w/ other test cleanup here: 202d18dfc7ca07ede7f41ed2f5abefcb9a2d0126

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
ben-taussig-solo: Resolved, w/ other test cleanup here: 202d18dfc7ca07ede7f41ed2f5abefcb9a2d0126

test/extensions/filters/http/transformation/inja_transformer_replace_test.cc
ben-taussig-solo: Resolved, w/ other test cleanup here: 202d18dfc7ca07ede7f41ed2f5abefcb9a2d0126

source/extensions/filters/http/transformation/inja_transformer.h
ben-taussig-solo: > as far as i can see, replaced_value_ is only ever written to, but never read

This is correct and intentional. The reason for this is that the various `ExtractionFunc` methods (`extractValue`, `replaceIndividualValue`, `replaceAllValues`) return `absl::string_view`, which is essentially a reference to a `std::string`. If we try to return a `absl::string` view from one of these methods without storing the post-replace value  somewhere, the underlying `std::string` will go out of scope before we can interact with the result

e.g., if we try (updated snippet from `replaceAllValues`)
```diff
  // If a match was found, replace all instances of the regex in the input value with the replacement_text_ value
  replaced = std::regex_replace(input, extract_regex_, replacement_text_.value());

-  // Store the replaced string in replaced_value_ and return an absl::string_view to it
-  replaced_value_ = std::move(replaced);
-  return absl::string_view(replaced_value_);
+  return absl::string_view(replaced);
```

we lose the value of `replaced` as soon as the method exits, causing the output to be interpreted as an empty string, even if that is not intended

ben-taussig-solo:
> does changing line 114 to extractor.set_regex(".*$") change anything? or maybe "^.*$"?

Yes -- either of those regexes results in normal/expected. The issue only exists when the regex is set to `.*`.

Note that this is caused by `std::regex_replace`, not by some quirk of our implementation. For a while I was thinking about introducing a check against this edge case to prevent this from happening, but recently I've been leaning more towards the side of leaving this in place, as I don't think it makes much sense to match an entire input when attempting to replace multiple values in the first place

source/extensions/filters/http/transformation/inja_transformer.h
ashishb-solo: okay, makes sense! could you just document that on the variable declaration?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -65,6 +67,29 @@ Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
...
+      if (!replacement_text_.has_value()) {
+        throw EnvoyException("REPLACE_ALL mode set but no replacement text provided");
+      }
+      if (group_ != 0) {
+        throw EnvoyException("REPLACE_ALL mode set but subgroup is not 0");
+      }
+      extraction_func_ = std::bind(&Extractor::replaceAllValues, this, _1, _2);
+      break;
+    default:
+      throw EnvoyException("Unknown mode");
```
ashishb-solo, line 91: this is probably fine, but you could also consider using `PANIC_DUE_TO_CORRUPT_ENUM` here

source/extensions/filters/http/transformation/inja_transformer.h
ashishb-solo: actually, i'm still a little hung up on this. why not change `ExtractionFunc` to return a string instead? or even better, a `std::unique_ptr<std::string>`? it feels a little strange to use a class variable to hold a return value on the stack.

overall, this seems like a slightly strange pattern to me, but if this is an accepted practice, i wouldn't get too hung up on it.

ben-taussig-solo: > Why not change `ExtractionFunc` to return a string instead

 - The `Extractor` uses the `ExtractionFunc`s in `Extractor::extract`, which predates this PR and returns an `absl::string_view`. If I wanted to update these methods to return a `std::string` or a `std::unique_ptr<std::string>`, I would need to either:
   - Update `Extractor::extract` to return an `std::string`(or `std::unique_ptr<std::string>`), which increases the risk (increased potential for string corruption bugs) and expands the scope of this PR
     - There really are a lot of changes that would need to follow from here if we want to go this route
   - Continue returning an `absl::string_view` from `Extractor::extract`, which leaves us in the same position we are now (i.e., we need to return a `absl::string_view` pointing to a `std::string`/`std::unique_ptr<std::string>` which will go out of scope when the function exits)

ben-taussig-solo: I agree, using a class variable to store the replace computation results isn't ideal. However, given how extractors currently interface with the transformation filter, it seems like the least disruptive way to handle this situation. We've got a series of steps outside of `Extractor::extract` that depend on `absl::string_view`s pointing to the replace/extract results. Tying the results to the lifetime of the `Extractor` object is a straightforward way to ensure that these `absl::string_view`s remain in scope.

jbohanon (approved):
:rock:

source/extensions/filters/http/transformation/inja_transformer.h
ashishb-solo: Thanks! I agree on all points. It's not ideal, but it works, and this isn't a great time to change this, so let's leave it for now. Perhaps we can consider revisiting this at some point in the future.

source/extensions/filters/http/transformation/inja_transformer.cc
ashishb-solo: > Can you link any issues or PRs you can recall where there's discussion about std::regex performance issues?

[Here's an issue in DLP](https://github.com/solo-io/solo-projects/issues/3957#). Similar situation - regexes are compiled at the time of configuration, but are still extremely slow. There are a few benchmarks linked in the issue that show how... hopeless `std::regex` can be.

Happy to leave this as is for now, but if this issue does bubble up, there's a bit of additional and (hopefully) helpful information on that thread

PR #305, opened 2024-02-08
[1.23] Feb 8 2024 CVE patches

jbohanon:


jbohanon:
It is possible that we will need to bump down bazel to 5.4.1 for this due to issues with downloading the prerelease and incompatibilities with 6.0.0 re: `@bazel_tools//platforms/` vs `@platforms//`

.bazelversion
```diff
@@ -1 +1 @@
-6.0.0-pre.20220421.3
\ No newline at end of file
+5.4.1
```
nfuden, line 1: Did we just have a mismatch in bazel versions?

jbohanon: So as far as I can tell the breaking change whereby `@bazel_tools//platforms` was removed occurred between the prerelease that we previously depended upon and `6.0.0`. `5.4.1` still has these tool definitions and the build still works

nfuden (approved):

PR #309, opened 2024-02-26
Add regex replace functionality to transformation filter extractors [Revised]

ben-taussig-solo:
## Description
 - This is a revised version of https://github.com/solo-io/envoy-gloo/pull/301 , which previously merged and was then reverted with https://github.com/solo-io/envoy-gloo/pull/302
 - The original PR was accidentally merged before OCTO had the opportunity to review, which is why we decided to revert
 - During the meeting in which we decided to revert #301, I committed to implementing some suggested changes:
   - *Addition of support for `RE2` as a supported regex engine in addition to C++ `std::regex`*
     - I did not end up doing this, for the following reasons:
       - This is a significant expansion of scope of this RFE. Adding support for this would require a significant time investment
       - RE2 is not designed for or performant relative to other regex engines when using capturing groups, which are central to the standard extraction and single value replace use cases. The goal of adding support for RE2 would be to improve the performance of extractor functionality, but there are other regex engines that are better suited for this use case
       - RE2 has differences in syntax and processing behavior that will require a significant amount of work to test effectively
   - *Support for dynamic metadata as an Extractor input*
     - ~~This has been added with this PR~~
     - Update: we did not end up including this functionality in the final PR
   - *Refactoring new replace all/replace individual value functionality to take and return `std::string` values, instead of `absl::string_view`*
     - This has been added in this PR

## API Changes
 - The `Extraction` (see Gloo Edge docs [here](https://docs.solo.io/gloo-edge/1.7.23/guides/traffic_management/request_processing/transformations/#extractors)) message has been updated to include 3 new fields:
   - `mode`
     - This is an enum which selects the mode of operation that selects the mode of operation that the extraction will use. Introduced with this PR are the `SINGLE_REPLACE` and `REPLACE_ALL` modes https://github.com/solo-io/envoy-gloo/blob/e2668be7d478800f32e0d8e9dd01ab16eafa75d5/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto#L157-L168 
   - `replacement_text`
     - when `SINGLE_REPLACE` or `REPLACE_ALL` modes are selected, the content matching the extractor's regex will be replaced with the value in `replacement_text`
   - `source.dynamic_metadata`
     -  ~~a `dynamic_metadata` field has been added to the `source` oneof, which specifies the source of the extraction. This value contains a key, the value of which in dynamic metadata will be provided to the extractor as input~~
     - Update: After discussion with Jacob, I have decide to remove this portion from the PR. I will put up a separate PR with the work that was done here. We decided that this functionality was too unstable to include in this PR

## Feature Overview 
 - This feature adds two new extractor "modes", which allow users to replace content in extracted values
   - These are `SINGLE_REPLACE` and `REPLACE_ALL` modes
   - In `SINGLE_REPLACE` mode, regex semantics are the same as with `EXTRACT` mode, i.e., the previous and now default extractor behavior. Users specify a regex that must match the entire input, and the content in the specified capture group is replaced with a configured string value (`replacement_text`)
   - in `REPLACE_ALL` mode, the regex is allowed to match the input multiple times. All matches are replaced by a configured string value (`replacement_text`)
 - This feature adds support for dynamic metadata as an extractor input
   - Dynamic metadata can be used as an input in all extraction modes


api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -153,24 +153,55 @@ message Transformation {
...
+    REPLACE_ALL = 2;
+  }
 
   // The source of the extraction
   oneof source {
     // Extract information from headers
     string header = 1;
     // Extract information from the request/response body
     google.protobuf.Empty body = 4;
+    // Extract information from dynamic metadata
```
ben-taussig-solo, line 176: I need to update this comment -- this value is used as a key in dynamic metadata and this comment needs to express that clearly

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -65,22 +80,111 @@ Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
...
+      }
+      return "";
+    }
+    default:
+      throw EnvoyException("Unsupported extraction source");
+  }
+}
+
+std::string
+Extractor::replace(Http::StreamFilterCallbacks &callbacks,
```
ben-taussig-solo, line 155: Look into whether it makes sense to extract shared code b/w this and `Extractor::extract`

ben-taussig-solo: Consider renaming replace to destructive extraction/extract destructive, etc.

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -545,16 +716,22 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
...
   if (advanced_templates_) {
     extractions.reserve(extractors_.size());
   }
 
   for (const auto &named_extractor : extractors_) {
     const std::string &name = named_extractor.first;
+    const ExtractionApi::Mode mode = named_extractor.second.mode();
     if (advanced_templates_) {
-      extractions[name] =
+      if (mode == ExtractionApi::REPLACE_ALL || mode == ExtractionApi::SINGLE_REPLACE) {
```
ben-taussig-solo, line 728: maybe just use a switch/case with all modes here?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -89,47 +102,89 @@ Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
...
+Extractor::getDynamicMetadataValue(Http::StreamFilterCallbacks &callbacks) const {
+  Upstream::ClusterInfoConstSharedPtr ci = callbacks.clusterInfo();
+  if (!ci) {
+    ENVOY_STREAM_LOG(debug, "No cluster info found for dynamic metadata", callbacks);
+    return absl::nullopt;
+  }
+
+  const envoy::config::core::v3::Metadata *cluster_metadata{};
+  cluster_metadata = &ci->metadata();
+  const ProtobufWkt::Value &value = Envoy::Config::Metadata::metadataValue(cluster_metadata, SoloHttpFilterNames::get().Transformation, dynamic_metadata_);
```
ben-taussig-solo, line 115: We probably don't want to be restricted to the Transformation filter namespace

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -89,47 +102,89 @@ Extractor::Extractor(const envoy::api::v2::filter::http::Extraction &extractor)
   }
 }
 
+absl::optional<std::string>
+Extractor::getDynamicMetadataValue(Http::StreamFilterCallbacks &callbacks) const {
+  Upstream::ClusterInfoConstSharedPtr ci = callbacks.clusterInfo();
```
ben-taussig-solo, line 107: `callbacks.streamInfo().dynamicMetdata()` might be better
Jacob mentioned that clusterInfo may be statically defined

ben-taussig-solo: Consider using setMetdata filter in e2e/integration testing

source/extensions/filters/http/transformation/inja_transformer.cc
nfuden: If we move to switch we can also use the CORRUPT_ENUM from upstream if need be

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
ben-taussig-solo: No longer relevant, as the functionality adding support for dynamic metadata extractor inputs has been removed from this PR

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: No longer relevant, as the functionality adding support for dynamic metadata extractor inputs has been removed from this PR

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: No longer relevant, as the functionality adding support for dynamic metadata extractor inputs has been removed from this PR

nfuden:
Step #0: Step #0 - "do_ci": [1mtest/extensions/filters/http/transformation/inja_transformer_test.cc:64:29: [0m[0;1;31merror: [0m[1mincompatible pointer types assigning to 'const std::unordered_map<std::string, std::string> *' (aka 'const unordered_map<basic_string<char, char_traits<char>, allocator<char>>, basic_string<char, char_traits<char>, allocator<char>>> *') from 'const std::unordered_map<std::string, absl::string_view> *' (aka 'const unordered_map<basic_string<char, char_traits<char>, allocator<char>>, basic_string_view<char>> *')[0m
Step #0: Step #0 - "do_ci":   typed_slot.extractions_ = &extractions;

ben-taussig-solo:
> Step #0: Step #0 - "do_ci": �[1mtest/extensions/filters/http/transformation/inja_transformer_test.cc:64:29: �[0m�[0;1;31merror: �[0m�[1mincompatible pointer types assigning to 'const std::unordered_map<std::string, std::string> *' (aka 'const unordered_map<basic_string<char, char_traits, allocator>, basic_string<char, char_traits, allocator>> *') from 'const std::unordered_map<std::string, absl::string_view> *' (aka 'const unordered_map<basic_string<char, char_traits, allocator>, basic_string_view> *')�[0m Step #0: Step #0 - "do_ci": typed_slot.extractions_ = &extractions;

resolved with 741ea24a34a8ecfce5c8b5a39f4bd13ab69bf71e

I had added a switch statement to select different modes of extractor operation, and left out `break;`s at the end of each case, causing us to always hit the default `PANIC_DUE_TO_CORRUPT_ENUM` case

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Updated here: https://github.com/solo-io/envoy-gloo/pull/309/commits/741ea24a34a8ecfce5c8b5a39f4bd13ab69bf71e

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -82,15 +83,24 @@ class Extractor : Logger::Loggable<Logger::Id::filter> {
...
+  std::string replaceIndividualValue(Http::StreamFilterCallbacks &callbacks,
+                                           absl::string_view value) const;
+  std::string replaceAllValues(Http::StreamFilterCallbacks &callbacks,
+                                     absl::string_view value) const;
 
   const Http::LowerCaseString headername_;
   const bool body_;
   const unsigned int group_;
   const std::regex extract_regex_;
+  const std::optional<std::string> replacement_text_;
```
jbohanon, line 102: For me or anyone else in the future wondering why we used `std::optional` over `absl::optional` (and what the differences are), `absl::optional`is now an [alias for the standard lib](https://groups.google.com/a/chromium.org/g/chromium-dev/c/nV0PlRJTg-U)

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -80,7 +102,34 @@ Extractor::extract(Http::StreamFilterCallbacks &callbacks,
     if (header_entries.empty()) {
       return "";
     }
-    return extractValue(callbacks, header_entries[0]->value().getStringView());
+    const auto &header_value = header_entries[0]->value().getStringView();
```
jbohanon, line 105: What is the purpose of pulling this out into a `const auto&` instead of inlining? I see that is done for the invocations of the newly added methods too

source/extensions/filters/http/transformation/inja_transformer.h
jbohanon, line 102: I think this could be a `const std::optional<const std::string&>` since the string is from the config object const ref passed into the constructor

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -80,7 +102,34 @@ Extractor::extract(Http::StreamFilterCallbacks &callbacks,
...
+    }
+  } else {
+    const Http::HeaderMap::GetResult header_entries = getHeader(header_map, headername_);
+    if (header_entries.empty()) {
+      return "";
+    }
+    const auto &header_value = header_entries[0]->value().getStringView();
+    if (mode_ == ExtractionApi::SINGLE_REPLACE) {
+      return replaceIndividualValue(callbacks, header_value);
+    } else {
```
jbohanon, line 130: It might be good to be explicit about our `mode_` expectation in the else case to guard against future API changes missing updating this code. In fact this should probably be a switch statement at that point

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -545,27 +661,50 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
...
 
   for (const auto &named_extractor : extractors_) {
     const std::string &name = named_extractor.first;
-    if (advanced_templates_) {
-      extractions[name] =
-          named_extractor.second.extract(callbacks, header_map, get_body);
-    } else {
-      absl::string_view name_to_split = name;
-      json *current = &json_body;
+    const ExtractionApi::Mode mode = named_extractor.second.mode();
```
jbohanon, line 671: Can this be inlined into the switch statement where it is used?

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -82,15 +83,24 @@ class Extractor : Logger::Loggable<Logger::Id::filter> {
   absl::string_view extract(Http::StreamFilterCallbacks &callbacks,
                             const Http::RequestOrResponseHeaderMap &header_map,
                             GetBodyFunc &body) const;
-
+  std::string extractDestructive(Http::StreamFilterCallbacks &callbacks,
+                      const Http::RequestOrResponseHeaderMap &header_map,
+                      GetBodyFunc &body) const;
+  ExtractionApi::Mode mode() const { return mode_; }
```
jbohanon, line 89:
```suggestion
  const ExtractionApi::Mode& mode() const { return mode_; }
```

jbohanon (commented):
Looking great!

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: I think this is left over from me trying to work around the issue of returning an `absl::string_view` while using the new destructive extraction methods. I'll switch it back to where it was before

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: It definitely should. I ran into this issue while writing unit tests and let it slide, so thanks for the catch!

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Anyway, updated that here: https://github.com/solo-io/envoy-gloo/pull/309/commits/5fcb54524240eac9fb9f56b609e1aa0d829513f4#diff-8c1e447dd62e766fbdd2f928d80d5332bdc2defbfd353a41f35db35fcd3b7deaR105

source/extensions/filters/http/transformation/inja_transformer.h
ben-taussig-solo: Updated here: https://github.com/solo-io/envoy-gloo/pull/309/commits/5fcb54524240eac9fb9f56b609e1aa0d829513f4#diff-7ac442c110145b36211bbfa1c31896337c06b25a47de63177a8943c48b380de8R102

I hit some compiler issues trying to use `const std::optional<const std::string&>`, but `const std::optional<const std::string>` works fine

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: done here: https://github.com/solo-io/envoy-gloo/pull/309/commits/5fcb54524240eac9fb9f56b609e1aa0d829513f4#diff-8c1e447dd62e766fbdd2f928d80d5332bdc2defbfd353a41f35db35fcd3b7deaR690

source/extensions/filters/http/transformation/inja_transformer.h
ben-taussig-solo: updated here: https://github.com/solo-io/envoy-gloo/pull/309/commits/5fcb54524240eac9fb9f56b609e1aa0d829513f4#diff-7ac442c110145b36211bbfa1c31896337c06b25a47de63177a8943c48b380de8R89

jbohanon (commented):
LGTM!

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -545,27 +665,49 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
     }
   }
   // get the extractions
-  std::unordered_map<std::string, absl::string_view> extractions;
+  std::unordered_map<std::string, std::string> extractions;
```
nfuden, line 668: Do we have any sense as to how much extra mem this will eat up? 
This should scale with number of extractions denoted right? So probably pretty low?

ben-taussig-solo: Yeah, this is an additional string per extraction, though to be fair that string could potentially be quite large if the input is a large request body or something like that.

ben-taussig-solo: `per_connection_buffer_limit_bytes` defaults to ~3.2kb (https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto), so I think it's reasonable to assume the upper limit for the size of this new string is around 10kb, with some caveats:
 - We should probably recommend that users don't use request/response bodies as inputs to extractors, especially if there is a chance that these bodies could be large
 - The size of these strings could grow larger if you made a serious mistake in configuration (i.e. write a regex that matches every character, then replace every match with a N-character string)
 - With the highlighted change, we are now converting the `absl::string_view` output of non-destructive extractions (i.e. the old extractor behavior) to `std::string` -- I'm currently looking into whether we can avoid this

nfuden (dismissed):

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -105,6 +158,73 @@ Extractor::extractValue(Http::StreamFilterCallbacks &callbacks,
...
+  // create an iterator to search for matches of extract_regex_ in the input string
+  std::sregex_iterator it(input.begin(), input.end(), extract_regex_);
+  std::sregex_iterator end_it; // default end iterator for comparison
+
+  // return an empty string if the regex doesn't match any part of the input value
+  if (it == end_it) {
+    ENVOY_STREAM_LOG(debug, "replaceAllValues: extractor regex did not match input", callbacks);
+    return "";
+  }
+
```
yuval-k, line 223: I see that this function returns the whole input, and not just the parts matched by the regex. 
If that's the case, we can keep this semantic even when the regex doesn't match, and remove this block?

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -162,15 +175,31 @@ message Extraction {
...
   // If your regex contains capturing groups, use this field to determine which
   // group should be selected.
+  // For EXTRACT and SINGLE_REPLACE, refers to the portion of the text
+  // to extract/replace.
+  // Config will be rejected if this is specified in REPLACE_ALL mode.
   uint32 subgroup = 3;
+
+  // The string to replace the matched portion of the source with.
+  // Used in SINGLE_REPLACE and REPLACE_ALL modes.
+  google.protobuf.StringValue replacement_text = 5;
```
yuval-k, line 198: this value goes directly into regex_replace. It has semantics that we should document. see the example here: https://en.cppreference.com/w/cpp/regex/regex_replace

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -82,15 +83,24 @@ class Extractor : Logger::Loggable<Logger::Id::filter> {
   absl::string_view extract(Http::StreamFilterCallbacks &callbacks,
                             const Http::RequestOrResponseHeaderMap &header_map,
                             GetBodyFunc &body) const;
-
+  std::string extractDestructive(Http::StreamFilterCallbacks &callbacks,
```
yuval-k, line 86: why is this function different? it seems to be semantically the same as extract? should we merge them to one function?
if we do that i think we can also get rid of mode() below

yuval-k: i see that they are only different by return value. but given that the string_view will be converted to string anyway in the transformation, there's no advantage in having extract return a string_view

ben-taussig-solo: I posted an update here which I think should make `extract` and `extractDestructive` meaningfully distinct: 9a543cf3abd968a8068bb11823c6c1958080fd13

Specifically, this creates two maps, `extractions` and `destructive_extractions` which have different signatures, which I believe should mean that the `string_view` returned from `extract` will not be converted to `std::string` anymore.

If I'm missing something and it still will be converted to `std::string`, then just let me combine these into one

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Removing this block causes this mode (REPLACE_ALL) to behave differently than the other modes in the case where the regex does not match the input

With EXTRACT (i.e. previous default extractor behavior) and SINGLE_REPLACE modes, if the regex does not match the input, we return an empty string. With the suggested change, if the regex does not match the input, we return the input.

Basically, I'm concerned that it could be confusing for users who are debugging problematic extractors to have behaviors vary between different modes.

I'm still open to revising and/or removing this block, but that's why it's in place

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
ben-taussig-solo: Updated this here: a4ed34d857bbefa91f2984df2ff57439cb79280a 
```proto
  // Used in SINGLE_REPLACE and REPLACE_ALL modes.
  // `replacement_text` is used to format the substitution for matched sequences in the input string
  // - In SINGLE_REPLACE mode, the content in the subgroup-th capturing group is replaced with the `replacement_text`.
  // - In REPLACE_ALL mode, each sequence matching the specified regex in the in the input is replaced with the `replacement_text`.
  //     The replacement_text may contain special syntax, such as $1, $2, etc., to refer to captured groups within the regular expression.
  //     The value contained within `replacement_text` is treated as a string, and is passed to std::regex_replace as the replacement string.
  //     see https://en.cppreference.com/w/cpp/regex/regex_replace for more details.
```

I also added a test demonstrating some of this functionality here: ef5b4d9f594199f07b4ac26141f1ff4b84d36232

source/extensions/filters/http/transformation/inja_transformer.cc
ben-taussig-solo: Updated this here: 6469b2614ffaaa98a82ccce921bdad77e89aacd9

Now, when the regex doesn't match in `REPLACE_ALL` mode, we return the input instead of an empty string

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -153,6 +153,19 @@ message Transformation {
 // Extractions can be used to extract information from the request/response.
 // The extracted information can then be referenced in template fields.
 message Extraction {
+  // The mode of operation for the extraction.
+  enum Mode {
+    // Default mode. Extract the value of the subgroup-th capturing group.
+    EXTRACT = 0; 
+    // Replace the value of the subgroup-th capturing group with the replacement_text.
+    // Note: replacement_text must be set for this mode.
+    SINGLE_REPLACE = 1;
```
nfuden, line 162: Shouldnt this act like replace_all

ben-taussig-solo: Updated `SINGLE_REPLACE` to return the original input if no match is found bd9380cac43b5f488c2161e6923d92aac1577983

I was hesitant about this at first, but this logic changed my mind:
 - in extract mode, a user's goal is to extract a value from a given request. if the regex doesn't match, then we can't extract that value, so we return nothing
 - in the replace modes, a user's goal is to overwrite portions of a given input. if the regex doesn't match, we can't overwrite anything, so we return the original input

I think it's also much easier for users to understand if the new modes have the same behavior when no match is found

nfuden (approved):

PR #310, opened 2024-02-28
tuneup extensions check

jbohanon:


ben-taussig-solo (approved):
These changes line up closely with the closed source companion PR (https://github.com/solo-io/envoy-gloo-ee/pull/740) and I've confirmed manually that the script:
 - runs quickly
 - passes with the current extensions_build_config.bzl (this is to be expected)
 - passes when extensions in the current extensions_build_config.bzl are commented out (again, expected behavior)
 - fails when extensions in the current extensions_build_config.bzl are removed from the file

PR #316, opened 2024-03-14
Use private pool cloud build worker

jbohanon:
These changes are part of the initiative to reduce flaking due to runners killing the processfor lack of memory. The investigation can be found [here](https://docs.google.com/document/d/1e-kpbV1mkSUnOiWFdn5xGcOfFZo-Sr4lcUIFhLmjpx0/edit?pli=1).

### DIFF
--- a/ci/cloudbuild.yaml
+++ b/ci/cloudbuild.yaml
@@ -63,7 +63,7 @@ artifacts:
     paths: ['linux/amd64/build_envoy_release/envoy']
 
 tags:
-  - "envoy-gloo"
+  - "repo_envoy-gloo"
 
 availableSecrets:
   inline:
--- a/ci/cloudbuild.yaml
+++ b/ci/cloudbuild.yaml
@@ -64,6 +64,11 @@ artifacts:
 
 tags:
   - "repo_envoy-gloo"
+  # This tag can be used to filter for or out jobs which are spawned by the main job
+  # submitted by build-bot. It's somewhat redundant as one could filter on `tags~^pr`
+  # to achieve the same effect since that tag is added to main jobs by build-bot,
+  # but this is somewhat less esoteric
+  - "sub-job"
 
 availableSecrets:
   inline:
--- a/ci/cloudbuild.yaml
+++ b/ci/cloudbuild.yaml
@@ -53,13 +53,17 @@ steps:
   - 'TAGGED_VERSION=$TAG_NAME'
 
 options:
-  machineType: 'N1_HIGHCPU_32'
+  pool:
+    name: 'projects/solo-public/locations/us-central1/workerPools/envoy-gloo-runner'
 timeout: 20000s
 
 artifacts:
   objects:
     location: 'gs://solo-public-artifacts.solo.io/envoy/$COMMIT_SHA/'
     paths: ['linux/amd64/build_envoy_release/envoy']
+
+tags:
+  - "envoy-gloo"
 
 availableSecrets:
   inline:
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -33,12 +33,8 @@ fi
 
 export ENVOY_SRCDIR=$SOURCE_DIR
 
-# google cloud build times out when using full throttle.
-# additionally, we see builds killed due to OOM at high concurrency.
-export NUM_CPUS=10
-
 # google cloud build doesn't like ipv6
-export BAZEL_EXTRA_TEST_OPTIONS="--test_env=ENVOY_IP_TEST_VERSIONS=v4only --test_output=errors --jobs=${NUM_CPUS}"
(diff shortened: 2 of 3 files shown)


cloudbuild.yaml
```diff
@@ -1,10 +1,12 @@
 steps:
 
 - name: 'gcr.io/cloud-builders/gcloud'
-  args: ['builds','submit','--config=ci/cloudbuild.yaml','--substitutions','TAG_NAME=$TAG_NAME,COMMIT_SHA=$COMMIT_SHA,_BUILD_TYPE=']
+  id: "standard"
+  args: ['builds','submit','--region=us-central1','--config=ci/cloudbuild.yaml','--substitutions','TAG_NAME=$TAG_NAME,COMMIT_SHA=$COMMIT_SHA,_BUILD_TYPE=']
 
 - name: 'gcr.io/cloud-builders/gcloud'
-  args: ['builds','submit','--config=ci/cloudbuild.yaml','--substitutions','COMMIT_SHA=$COMMIT_SHA,_BUILD_TYPE=clang-asan']
+  id: "asan"
```
ben-taussig-solo, line 8: thanks for this

ci/cloudbuild.yaml
```diff
@@ -53,14 +53,23 @@ steps:
...
+    name: 'projects/solo-public/locations/us-central1/workerPools/envoy-gloo-runner'
 timeout: 20000s
 
 artifacts:
   objects:
     location: 'gs://solo-public-artifacts.solo.io/envoy/$COMMIT_SHA/'
     paths: ['linux/amd64/build_envoy_release/envoy']
 
+tags:
+  - "repo_envoy-gloo"
```
ben-taussig-solo, line 66: Should we add a tag like `using-private-pool` so we can compare jobs run with/without private runners? I haven't spent much time using tags/filters on jobs in gcloud so I'm not sure whether that info is easily available

cloudbuild.yaml
```diff
@@ -1,10 +1,12 @@
 steps:
 
 - name: 'gcr.io/cloud-builders/gcloud'
-  args: ['builds','submit','--config=ci/cloudbuild.yaml','--substitutions','TAG_NAME=$TAG_NAME,COMMIT_SHA=$COMMIT_SHA,_BUILD_TYPE=']
+  id: "standard"
+  args: ['builds','submit','--region=us-central1','--config=ci/cloudbuild.yaml','--substitutions','TAG_NAME=$TAG_NAME,COMMIT_SHA=$COMMIT_SHA,_BUILD_TYPE=']
```
ben-taussig-solo, line 5: Are we interested in adding either the `repo_envoy-gloo` or `main-job` tags on these jobs? Should we include the build type or any other information as a tag?

jbohanon: `repo_envoy-gloo` is added by build bot [here](https://github.com/solo-io/build-bot/blob/master/pkg/internal/build/build.go#L115).

I could see value in adding `main-job` but its omission is also good enough for me. Happy to add it though

ci/do_ci.sh
```diff
@@ -33,12 +33,8 @@ fi
 
 export ENVOY_SRCDIR=$SOURCE_DIR
 
-# google cloud build times out when using full throttle.
-# additionally, we see builds killed due to OOM at high concurrency.
-export NUM_CPUS=10
```
ben-taussig-solo, line 38: How stable is CI if we don't restrict the number of processes? Should we expect to see memory issues when we end up doing the 1.30 bump?

ci/cloudbuild.yaml
jbohanon: This is accessible in the output of `gcloud builds describe <id>`. It is `{}` for a default pool run and non-empty for a private pool run. Additionally, it is a sort/filter option in the UI

ci/do_ci.sh
jbohanon: I do not expect to see memory issues again since the runner we are using has such a vastly greater ratio of memory:cores. It's not so much about the number of processes, it's more an issue of how much memory is available for each process.

ci/cloudbuild.yaml
ben-taussig-solo: thanks! resolving the thread

cloudbuild.yaml
ben-taussig-solo: given that there are only two job types, `main-job` doesn't seem necessary. resolving the thread.

ci/do_ci.sh
ben-taussig-solo: makes sense, thanks!

ben-taussig-solo (approved):

jbohanon:
paging Mr. Dozer...

PR #317, opened 2024-03-14
add action for static analysis

jbohanon:
This includes Inja static analysis fix from [this PR](https://github.com/solo-io/inja/pull/7) and Protobuf static analysis fix from [this PR](https://github.com/solo-io/envoy-fork/pull/20)

### DIFF
--- a/bazel/repository_locations.bzl
+++ b/bazel/repository_locations.bzl
@@ -1,14 +1,16 @@
 REPOSITORY_LOCATIONS = dict(
     envoy = dict(
         # envoy 1.29.2 with backported ext_proc updates
-        commit = "b5c77750dc0f9de19478c89776f580fb9336fc99",
+        # also includes libprotobuf patch for dead stores
+        commit = "4215f08044ae0bab9e74a8b248bb6480709a3a1f",
         remote = "https://github.com/solo-io/envoy-fork",
     ),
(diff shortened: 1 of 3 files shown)


bazel/repository_locations.bzl
```diff
@@ -1,7 +1,8 @@
 REPOSITORY_LOCATIONS = dict(
+    # can't have more than one comment between envoy line and commit line in
```
ben-taussig-solo, line 2: The `-A 2` on this line
https://github.com/solo-io/envoy-gloo/blob/46ca9b475584b21655fb40c617769ea870c35838/ci/check_extensions_build_config.sh#L27

Selects how many lines to search for to find the envoy commit

jbohanon: Yeah. We can't arbitrarily bump it though because eventually it will be looking at the block for the next dependency right? Unless the envoy dependency is just always listed last but that seems just as arbitrary to me

ben-taussig-solo: > We can't arbitrarily bump it though because eventually it will be looking at the block for the next dependency right?

yes, that is correct.

ben-taussig-solo (approved):
Looks good to me! Glad to have static analysis up and running

PR #328, opened 2024-04-11
Fix: Lambda json parsing

ben-taussig-solo:
# Description

## Summary of bug
 - The bug occurs when unwrapAsApiGateway is enabled in the lambda filter
   - in this case, we attempt to convert a json payload returned by the upstream lambda into a response forwarded to the downstream client
 - in this case, we assumed that we'd receive a payload w/ the form:
```json
{
  "multiValueHeaders": {
    "x-header-foo": [
      "bar1",
       "bar2"
    ]
  }
}
```

 - instead, we choked on a payload like this (value inside inner array has unexpected type)
```json
{
  "multiValueHeaders": [
    "x-header-foo": [
      null
    ]
  ]
  
}
```

### Primary fix
 - when iterating over the header values for a given key in the multiValueHeaders object, we now explicitly confirm that the value is a string before assigning it to a variable of type `std::string`
```diff
- const auto& header_value = inner_it.value().get<std::string>();
+ std::string header_value;
+ if (inner_it.value().is_string()) {
+   header_value = inner_it.value().get<std::string>();
+ } else {
+   header_value = inner_it.value().dump();
+ }
```
### related fix
 - this PR also validates that the value pointed to by a given header name in the multiValueHeaders object is an array before attempting to iterate over it
 - specifically, we reject the request if the value is of type object 
 - singleton types are processed without issue, so we do not need to reject requests which contain these types 
```cpp
// need to validate that header_values is an array/iterable
if (!header_values.is_array()) {
  // if it's an object, we reject the request
  if (header_values.is_object()) {
    ENVOY_STREAM_LOG(debug, "invalid multi header value object", stream_filter_callbacks);
    ApiGatewayError error = {500, "500", "invalid multi header value object"};
    return ApiGatewayTransformer::format_error(*response_headers, body, error, stream_filter_callbacks);
  }

  ENVOY_STREAM_LOG(debug, "warning: using non-array value for multi header value", stream_filter_callbacks);
}
```
### status code type checking
 - this PR also adds some additional type checking around the status field object
 - in particular, we remove an unecessary try/catch block, and get stricter about negative/float/non-integer status code values
```diff
- try {
-   status_value = json_body["statusCode"].get<uint64_t>();
- } catch (std::exception& exception){
-   ENVOY_STREAM_LOG(debug, "Error parsing statusCode: " + std::string(exception.what()), stream_filter_callbacks);
-   ApiGatewayError error = {500, "500", "Non-integer status code"};
+ if (!json_body["statusCode"].is_number_unsigned()) {
+   // add duplicate log line to not break tests for now
+   ENVOY_STREAM_LOG(debug, "cannot parse non unsigned integer status code", stream_filter_callbacks);
+   ENVOY_STREAM_LOG(debug, "received status code with value: " + json_body["statusCode"].dump(), stream_filter_callbacks);
+   ApiGatewayError error = {500, "500", "cannot parse non unsigned integer status code"};
  return ApiGatewayTransformer::format_error(*response_headers, body, error, stream_filter_callbacks);
}
+ status_value = json_body["statusCode"].get<uint64_t>();
```


source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
```diff
@@ -132,17 +134,32 @@ void ApiGatewayTransformer::transform_response(
...
+
+          ENVOY_STREAM_LOG(debug, "warning: using non-array value for multi header value", stream_filter_callbacks);
+        }
         for (json::const_iterator inner_it = header_values.cbegin(); inner_it != header_values.cend(); inner_it++) {
-          const auto& header_value = inner_it.value().get<std::string>();
+          std::string header_value;
+          if (inner_it.value().is_string()) {
+            header_value = inner_it.value().get<std::string>();
+          } else {
+            header_value = inner_it.value().dump();
```
ben-taussig-solo, line 161: can we remove this conditional and just use .dump() -- does .dump() change string formatting?

test/extensions/transformers/aws_lambda/transformer_test.cc
```diff
@@ -3,13 +3,8 @@
 #include "source/extensions/transformers/aws_lambda/api_gateway_transformer.h"
 
 #include "test/mocks/http/mocks.h"
-#include "test/mocks/server/mocks.h"
-#include "test/test_common/utility.h"
 
-#include "fmt/format.h"
 #include "nlohmann/json.hpp"
-#include "gmock/gmock.h"
-#include "gtest/gtest.h"
```
ben-taussig-solo, line 12: removed (along with the changes to the BUILD file) to improve compilation times for these tests

on my machine, this reduced the time to compile small changes to the already-cached suite from 80s to 60s

ben-taussig-solo:
/kick looks like the thing where the build is running in the background but status is not reported correctly via github.

source/extensions/transformers/aws_lambda/api_gateway_transformer.cc
ben-taussig-solo: unfortunately this will not work for our purposes -- the "dumped" string value has escaped quotes, e.g. `\"test-header-value\"`

the documentation for `dump()` is [here](https://json.nlohmann.me/api/basic_json/dump/). From what I can tell, there isn't a setting that we can pass to `dump()` to work around this

test/extensions/transformers/aws_lambda/transformer_test.cc
```diff
@@ -305,7 +300,296 @@ TEST(ApiGatewayTransformer, error) {
...
+                                         {":authority", "www.solo.io"},
+                                         {"x-test", "789"},
+                                         {":path", "/users/123"}};
+  Http::TestResponseHeaderMapImpl response_headers{};
+  ApiGatewayTransformer transformer;
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> filter_callbacks_{};
+
+  if (expected_error_message.empty()) {
+    transformer.transform(response_headers, &headers, *bodyPtr, filter_callbacks_);
+    auto lowercase_multi_header_name = Http::LowerCaseString("test-multi-header");
```
ashishb-solo, line 321: nit: would be nice to move `"test-multi-header"` to an external `const` value

test/extensions/transformers/aws_lambda/transformer_test.cc
```diff
@@ -305,7 +300,296 @@ TEST(ApiGatewayTransformer, error) {
...
+    std::string expected_multi_value_header = "multi-value-0,multi-value-1,multi-value-2") {
+  Http::TestRequestHeaderMapImpl headers{{":method", "GET"},
+                                         {":authority", "www.solo.io"},
+                                         {"x-test", "789"},
+                                         {":path", "/users/123"}};
+  Http::TestResponseHeaderMapImpl response_headers{};
+  ApiGatewayTransformer transformer;
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> filter_callbacks_{};
+
+  if (expected_error_message.empty()) {
```
ashishb-solo, line 319: nit: would prefer to use `std::optional` instead of empty strings as sentinel values

test/extensions/transformers/aws_lambda/transformer_test.cc
```diff
@@ -305,7 +300,296 @@ TEST(ApiGatewayTransformer, error) {
...
 }
 
+// helper used in multi value headers type safety tests
+// - bodyPtr: json payload in the format used by AWS API Gateway/returned from an upstream Lambda
+// - expected_error_message: if present, expect that an error message will be logged that contains this string
+// - expected_multi_value_header: if present, expect that the first value of the 'test-multi-header' header in the response will be equal to this string,
+void test_multi_value_headers_transformation(
+    std::unique_ptr<Buffer::OwnedImpl> bodyPtr,
+    std::string expected_error_message = "Error transforming response: [json.exception.type_error.302] type must be string, but is null",
+    std::string expected_multi_value_header = "multi-value-0,multi-value-1,multi-value-2") {
```
ashishb-solo, line 310: nit: would prefer to see these as external `const` values

nfuden (approved):

PR #330, opened 2024-04-19
Add base64url fallback to base64_decode

jbohanon:
JWTs and other usages use a modified character set from the standard base64. This variant is called base64url. When our Transformation filter is presented with a base64url encoded string to the base64_decode inja callback, the transformation fails silently because the decode returned an empty string.

This PR introduces a new pair of callbacks capable of handling `base64url` encoding

### DIFF
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -239,8 +239,14 @@ TransformerInstance::TransformerInstance(ThreadLocal::Slot &tls, Envoy::Random::
   env_.add_callback("base64_encode", 1, [this](Arguments &args) {
     return base64_encode_callback(args);
   });
+  env_.add_callback("base64url_encode", 1, [this](Arguments &args) {
+    return base64url_encode_callback(args);
+  });
   env_.add_callback("base64_decode", 1, [this](Arguments &args) {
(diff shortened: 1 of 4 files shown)


source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -390,7 +391,16 @@ json TransformerInstance::base64_encode_callback(const inja::Arguments &args) co
 
 json TransformerInstance::base64_decode_callback(const inja::Arguments &args) const {
```
nfuden, line 392: Should we also introduce a callback that is strict base64?
Also can we add a comment with a link to the rfc? https://datatracker.ietf.org/doc/html/rfc4648#section-5

jbohanon: good call added.

I was originally trying to surgically fix the problem, but the add wasn't a big lift.

A few tests would be useful but I really don't have time right now.

yuval-k: I would prefer to keep current impl as is to avoid changing behavior..m

jbohanon: I have reverted the change to the current impl so we only have net-new. The test now tests the `base64url_decode` callback also

PR #337, opened 2024-05-09
1.27: Add base64url fallback to base64_decode (#330)

nfuden:
Support for base64url  in transformations


test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -882,6 +882,28 @@ TEST_F(InjaTransformerTest, Base64DecodeTestString) {
...
+
+  InjaTransformer transformer(transformation, rng_, google::protobuf::BoolValue(), tls_);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+
+  Buffer::OwnedImpl body("");
+  transformer.transform(headers, &headers, body, callbacks);
+  EXPECT_NE(std::string::npos, body.toString().find(expect_string));
+}
+
```
ashishb-solo, line 906: would it be possible to add a test where we test for equality of the base64-decode callback?

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -66,7 +66,9 @@ class TransformerInstance {
   nlohmann::json env(const inja::Arguments &args) const;
   nlohmann::json cluster_metadata_callback(const inja::Arguments &args) const;
   nlohmann::json base64_encode_callback(const inja::Arguments &args) const;
+  nlohmann::json base64url_encode_callback(const inja::Arguments &args) const;
```
ashishb-solo, line 69: can we also add a test for base64url_encode?

changelog/v1.27.5-patch2/base64url.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+- type: FIX
+  issueLink: https://github.com/solo-io/envoy-gloo/issues/329
+  resolvesIssue: false
+  description: >-
+    Fall back to attempting to decode Base64Url if Base64 decoding fails in the Inja transformation
```
ashishb-solo, line 6: i'm a little confused by the language here. where exactly does the "fall back" occur when decoding base64 urls? as far as i can see, we are only adding 2 new functions (base64url_encode/decode), but i don't see where we fall back to using it if failure occurs in the regular base64 encoding/decoding

ashishb-solo, line 6: according to the linked issue, "the transformation fails silently because the decode returned an empty string" - while we're at this, is it possible to add a warning of some sort when this happens?

ashishb-solo (commented):
a few comments, happy to discuss if it would be helpful

changelog/v1.27.5-patch2/base64url.yaml
nfuden: I think thats out of scope for this pr as its a problem with many of our transformations and has been in general the accepted state.

Some customers may rely on this today

nfuden, line 6:
```suggestion
    Expose new Base64Url aware transformations that will fall back to attempting to decode Base64Url if Base64 decoding fails in the Inja transformation. These are new options in case users are relying on current behavior and the strict spec for base64
```

changelog/v1.27.5-patch2/base64url.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
+- type: FIX
+  issueLink: https://github.com/solo-io/envoy-gloo/issues/329
+  resolvesIssue: false
+  description: >-
+    Expose new Base64Url aware transformations that will fall back to attempting to decode Base64Url if Base64 decoding fails in the Inja transformation. These are new options in case users are relying on current behavior and the strict spec for base64
+    base64_decode callback. This can happen especially when attempting to decode JWTs.
```
ashishb-solo, line 7:
```suggestion
    Expose `base64url_encode` and `base64url_decode` Inja transformers, which are new transformations that will fall back to attempting to decode Base64Url if Base64 decoding fails. These are provided as new options, and the existing ones remain unchanged to preserve backward compatibility in case users are relying on current behavior. This can happen especially when attempting to decode JWTs.
```

PR #340, opened 2024-05-23
upstream transformation_filter

EItanya:
This allows the transformation filter to be run as an `upstream_http_filter`. This means it can be set either in the cluster or the router filter.

As a part of this work I also needed to add an additional upstream_http_filter which I have name the `WaitFilter`. This is an unfortunate side effect of how upstream selection works in relation to transformations which require modifying body and headers. Upstream selection in envoy happens async and then fires off a callback when it's done. The `WaitFilter` [hooks into that callback](https://github.com/solo-io/envoy-gloo/pull/340/commits/b5f84d3b6ae1f258b23ab3813132afdfac9bfc54#diff-ffd3d9d6082340a7bde0d3d0f3f2a6d421be280db50b8113a6e90addf23eb67fR37-R39), stopping all iteration until it's called, and then resumes filter iteration. This filter MUST be added when modifying the body as a part of an upstream http_filter. If we decide to also include lambda functionality this may also be necessary.
BOT NOTES: 
resolves https://github.com/solo-io/envoy-gloo/issues/342


test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -967,6 +967,23 @@ TEST_F(InjaTransformerTest, SubstringTwoArguments) {
...
+  TransformationTemplate transformation;
+
+  transformation.mutable_body()->set_text("{{word_count(body())}}");
+  transformation.set_parse_body_behavior(TransformationTemplate::DontParse);
+
+  InjaTransformer transformer(transformation, rng_, google::protobuf::BoolValue(), tls_);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+
+  auto test_string = "why don't you accept me";
```
nfuden, line 981: nit: we accept you

source/extensions/filters/http/solo_well_known_names.h
```diff
@@ -26,6 +26,8 @@ class SoloHttpFilterNameValues {
   // Transformation filter
   // TODO(talnordan): Consider "io.solo.filters.http.transformation".
   const std::string Transformation = "io.solo.transformation";
+  // Wait filter
+  const std::string Wait = "io.solo.wait";
```
nfuden, line 30:
```suggestion
  const std::string Wait = "io.solo.filters.http.wait";
```

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -236,6 +236,15 @@ TransformerInstance::TransformerInstance(ThreadLocal::Slot &tls, Envoy::Random::
...
   });
+  env_.add_callback("cluster_metadata", 1, [this](Arguments &args) {
+    return cluster_metadata_callback(args);
+  });
+  env_.add_callback("dynamic_metadata", 1, [this](Arguments &args) {
+    return dynamic_metadata_callback(args);
+  });
+  env_.add_callback("host_metadata", 1, [this](Arguments &args) {
+    return host_metadata_callback(args);
+  });
```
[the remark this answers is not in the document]
EItanya: I'm not so sure what you mean

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -410,6 +414,44 @@ json TransformerInstance::base64url_decode_callback(const inja::Arguments &args)
...
+int TransformerInstance::word_count(const nlohmann::json* input)  {
+  if (input->is_string()) {
+    const std::string &input_string = input->get_ref<const std::string &>();
+    unsigned long ctr = 0; // Initializing a counter variable to count words
+
+    // Loop through the string and count spaces to determine words
+    for (unsigned long x = 0; x < input_string.length(); x++) {
+      if (input_string[x] == ' ') // Checking for spaces to count words
+        ctr++; // Increment the counter for each space found
+    }
```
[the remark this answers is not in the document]
EItanya: Yea, this bit needs some work, definitely would love some suggestions :)

source/extensions/filters/http/transformation/inja_transformer.cc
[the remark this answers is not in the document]
nfuden: instead of "dynamic_metadata" as a callback
it might make sense to be: "metadata"

Or do something where we can hook into upstream's command operator framework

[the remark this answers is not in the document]
EItanya: Can you send me some examples? Didn't you say upstream's metadata impl is very un performant?

EItanya:
@nfuden I split the word_count and other functions into a separate PR for simplicity

source/extensions/filters/http/transformation/inja_transformer.cc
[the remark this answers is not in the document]
nfuden: Oh the current approach is especially as it gets large. Thats actually a good point not to do it. Closing but for reference the header manipulation delegates to substitution formatter which is how it gets its resolution https://github.com/envoyproxy/envoy/blob/release/v1.30/source/common/router/header_parser.cc#L48

source/extensions/filters/http/transformation/inja_transformer.cc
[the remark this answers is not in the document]
EItanya: Moved this into another PR

source/extensions/filters/http/wait/filter.h
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  // UpstreamCallbacks
+  void onUpstreamConnectionEstablished() override;
+
+private:
+  Http::StreamDecoderFilterCallbacks *decoder_callbacks_{};
+  // Determines whether the stream has been ended for running the filter in upstream mode.
+  absl::optional<bool> latched_end_stream_;
```
yuval-k, line 48: from the code it looks like this just needs to be a bool vs an optional<bool> (i.e. the value of the bool is not checked; just that of the optional)?

also, it seems to be used to indicate that iteration was paused to wait for the upstream connection. maybe `paused_iteration_` would be a better name?

source/extensions/filters/http/wait/filter.h
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  Http::FilterTrailersStatus decodeTrailers(Http::RequestTrailerMap&) override {
+    return Http::FilterTrailersStatus::Continue;
+  }
+
+  void setDecoderFilterCallbacks(
+      Http::StreamDecoderFilterCallbacks &callbacks) override {
+    decoder_callbacks_ = &callbacks;
+    if (callbacks.upstreamCallbacks()) {
```
yuval-k, line 37: should we check `decoder_callbacks_->upstreamCallbacks()->upstream()` is not nil here, and skip adding us to the callbacks?

source/extensions/filters/http/transformation/transformation_filter.h
```diff
@@ -122,6 +122,8 @@ class TransformationFilter : public Http::StreamFilter,
   bool destroyed_{};
 
   FilterConfigSharedPtr filter_config_;
+  // Determines whether the stream has been ended for running the filter in upstream mode.
+  absl::optional<bool> latched_end_stream_;
```
yuval-k, line 126: this moved to wait filter, right? delete?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -780,10 +799,21 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
...
+      // Need to check if number
+      try {
+        const auto int_output = std::stoi(output);
+        ProtobufWkt::Struct struct_obj;
+        ProtobufWkt::Value val;
+        val.set_number_value(int_output);
+        (*struct_obj.mutable_fields())[templated_dynamic_metadata.key_] = val;
+        callbacks.streamInfo().setDynamicMetadata(
+            templated_dynamic_metadata.namespace_, struct_obj);
+      } catch (...) { // This function can return multiple exceptions
```
yuval-k, line 811: what here can throw an exception? is there a way to avoid it?
the problem with `catch(...)` is that we have no idea what went wrong.. this will be hard to debug.

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -780,10 +799,21 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
   for (const auto &templated_dynamic_metadata : dynamic_metadata_) {
     std::string output = instance_->render(templated_dynamic_metadata.template_);
     if (!output.empty()) {
-      ProtobufWkt::Struct strct(
-          MessageUtil::keyValueStruct(templated_dynamic_metadata.key_, output));
-      callbacks.streamInfo().setDynamicMetadata(
-          templated_dynamic_metadata.namespace_, strct);
+      // Need to check if number
```
yuval-k, line 802: this breaks backwards compatbility; if customers have number like strings today, we will now change them to a number implicity. is there a way to explicitly ask for a number?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -764,6 +776,13 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
   typed_tls_data.context_ = &json_body;
   typed_tls_data.environ_ = &environ_;
   typed_tls_data.cluster_metadata_ = cluster_metadata;
+  typed_tls_data.dynamic_metadata_ = dynamic_metadata;
+
+  // If there is a value we're in a upstream filter
+  if (callbacks.upstreamCallbacks().has_value()) {
+    auto &upstream_callbacks = callbacks.upstreamCallbacks().value().get();
+    typed_tls_data.endpoint_metadata_ = upstream_callbacks.upstreamStreamInfo().upstreamInfo()->upstreamHost()->metadata();
+  }
```
yuval-k, line 785: check if `upstreamHost` is null here (in case we don't setup the wait filter in the control plane)

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -764,6 +776,13 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
   typed_tls_data.context_ = &json_body;
   typed_tls_data.environ_ = &environ_;
   typed_tls_data.cluster_metadata_ = cluster_metadata;
+  typed_tls_data.dynamic_metadata_ = dynamic_metadata;
+
+  // If there is a value we're in a upstream filter
+  if (callbacks.upstreamCallbacks().has_value()) {
```
yuval-k, line 782: `typed_tls_data.endpoint_metadata_` **must** be set to nil if not set to upstrema host metadata; as otherwise endpoint_metadata_ may "leak" 

maybe follow the example of how `dynamic_metadata` is set? create a variable that's nill; try to fill it, and in the end set it anyway in the typed_tls_data? this guarantees that it is always set to correct value

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -752,6 +753,17 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
...
 
+  // get cluster metadata
+  const envoy::config::core::v3::Metadata *dynamic_metadata{};
+  dynamic_metadata = &callbacks.streamInfo().dynamicMetadata();
+  // if (ci.get()) {
+  //   cluster_metadata = &ci->metadata();
+  // }
+
+  // const envoy::config::core::v3::Metadata *dynamic_metadata{};
+  // callbacks.upstreamCallbacks()
```
yuval-k, line 764: remove?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -343,45 +374,15 @@ json TransformerInstance::cluster_metadata_callback(const inja::Arguments &args)
     return stringval;
     break;
   }
+  case ProtobufWkt::Value::kStructValue:
+    ABSL_FALLTHROUGH_INTENDED;
```
yuval-k, line 378: why use fallthrough here and not separate the list and struct cases?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -343,45 +374,15 @@ json TransformerInstance::cluster_metadata_callback(const inja::Arguments &args)
...
-
-    addValue();
-
-    for (it++; it != listval.end(); it++) {
-      ss << ",";
-      addValue();
-    }
-    return ss.str();
+    std::string output;
+    auto status = value.has_struct_value()
```
yuval-k, line 381: is status unused?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -343,45 +374,15 @@ json TransformerInstance::cluster_metadata_callback(const inja::Arguments &args)
...
+                      ? ProtobufUtil::MessageToJsonString(value.struct_value(), &output)
```
yuval-k, line 382: do we have tests to confirm that the final value did not change?

yuval-k (commented):
left some comments. notes:
- add unit tests for new transformation functions.
- can we add an integration test of clearing route cache set in the transform in an upstream filter? (to make sure envoy doens't crash).
- this can also test the wait filter.

source/extensions/filters/http/wait/filter.h
EItanya: That's what this does, if upstreamCallbacks is set this is an upstream filter. Also at construction Upstream won't be available afaik

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: stoi can throw 2 exceptions, neither are super helpful here.

https://en.cppreference.com/w/cpp/string/basic_string/stol

EItanya: I can catch them if you want for better logging

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: What do you mean? For backwards compatibility? I thought you said no one uses this callback

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: I got rid of this completely

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: This was copy/pasta'd from the upstream JWT filter, I have no strong feeling

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: Yes, but the compiler requires it to be set anyway

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: Hmm, right now you we always use them together in the `upstream_context`, do we want to plan for it not being there if the control-plane never allows that situation?

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: Sounds good, will update

source/extensions/filters/http/transformation/inja_transformer.cc
yuval-k: hmm i'm not sure what you mean; but yea this might not be needed. let's chat about this

source/extensions/filters/http/transformation/inja_transformer.cc
yuval-k: idk how we might change the control plane in the future.. i think it's better to verify here

source/extensions/filters/http/wait/filter.h
yuval-k: even if we re-use an existing connection?

test/integration/transformation_filter_integration_test.cc
```diff
@@ -58,6 +58,16 @@ const std::string BODY_TRANSFORMATION =
         text: "{{abc}}"
 )EOF";
 
+const std::string BODY_TRANSFORMATION_METADATA =
+    R"EOF(
+  clear_route_cache: true
```
yuval-k, line 63: i would mention here and in the proto-docs that `clear_route_cache` does nothing for upstream filters, and is only set here to test that envoy behaves well even if it is set

test/integration/transformation_filter_integration_test.cc
```diff
@@ -153,10 +163,22 @@ class TransformationFilterIntegrationTest
...
+      HttpFilter filter;
+      filter.set_name(
+          Extensions::HttpFilters::SoloHttpFilterNames::get().Wait);
+      config_helper_.prependFilter(MessageUtil::getJsonStringFromMessageOrError(filter), downstream_filter_);
+      addEndpointMeta();
+    }
 
     if (transformation_string_ != "") {
+      // // set the default transformation
+      // config_helper_.addFilter(default_filter);
```
yuval-k, line 181: why is this commented out? should this be removed?

yuval-k (dismissed):

PR #341, opened 2024-05-24
Transformation word_count

EItanya:
Add a word_count function to the inja transformer. It can handle nested json.
BOT NOTES: 
resolves https://github.com/solo-io/envoy-gloo/issues/343

### DIFF
--- a/source/extensions/filters/http/transformation/inja_transformer.cc
+++ b/source/extensions/filters/http/transformation/inja_transformer.cc
@@ -236,15 +236,6 @@ TransformerInstance::TransformerInstance(ThreadLocal::Slot &tls, Envoy::Random::
   env_.add_callback("clusterMetadata", 1, [this](Arguments &args) {
     return cluster_metadata_callback(args);
   });
-  env_.add_callback("cluster_metadata", 1, [this](Arguments &args) {
-    return cluster_metadata_callback(args);
-  });
-  env_.add_callback("dynamic_metadata", 1, [this](Arguments &args) {
-    return dynamic_metadata_callback(args);
-  });
-  env_.add_callback("host_metadata", 1, [this](Arguments &args) {
-    return host_metadata_callback(args);
-  });
   env_.add_callback("base64_encode", 1, [this](Arguments &args) {
     return base64_encode_callback(args);
   });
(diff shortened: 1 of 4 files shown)


source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -410,6 +413,53 @@ json TransformerInstance::base64url_decode_callback(const inja::Arguments &args)
...
+    const auto &input_array = input->get_ref<const std::vector<json> &>();
+    for (auto & element : input_array) {
+      total_word_count += word_count(&element);
+    }
+    return total_word_count;
+  } else if (input->is_object()) {
+    int total_word_count = 0;
+    const auto element_obj = input->get_ref<const json::object_t &>();
+    for (auto & [key, value] : element_obj) {
+      // Key is always a string
```
yuval-k, line 451: can the key have multiple words?

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -967,6 +967,49 @@ TEST_F(InjaTransformerTest, SubstringTwoArguments) {
...
+
+  EXPECT_CALL(callbacks.stream_info_,
+              setDynamicMetadata(SoloHttpFilterNames::get().Transformation, _))
+      .Times(1)
+      .WillOnce(
+          Invoke([](const std::string &, const ProtobufWkt::Struct &value) {
+            auto field = value.fields().at("foo");
+            EXPECT_EQ(field.string_value(), "12");
+          }));
+  auto test_string = "{\"a\": \"Hal, what's the meaning of life?\", \"b\": \"I don't know John\"}";
```
yuval-k, line 1008: add a test case that confirms that multiple spaces don't effect word count?
do we want to test tabs too?

source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -39,6 +39,8 @@ struct ThreadLocalTransformerContext : public ThreadLocal::ThreadLocalObject {
   const nlohmann::json *context_;
   const std::unordered_map<std::string, std::string> *environ_;
   const envoy::config::core::v3::Metadata *cluster_metadata_;
+  Envoy::Upstream::MetadataConstSharedPtr endpoint_metadata_;
+  const envoy::config::core::v3::Metadata *dynamic_metadata_;
```
yuval-k, line 43: remove?

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: Yes they can! oops

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -410,6 +413,75 @@ json TransformerInstance::base64url_decode_callback(const inja::Arguments &args)
   return Base64Url::decode(input);
 }
 
+json TransformerInstance::word_count_callback(const inja::Arguments &args) {
+  return std::to_string(json_word_count(args.at(0)));
```
yuval-k, line 417: do we need the to_string() here? i.e. is there a disadvantage of returning an int?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -410,6 +413,75 @@ json TransformerInstance::base64url_decode_callback(const inja::Arguments &args)
...
+    const auto element_obj = input->get_ref<const json::object_t &>();
+    for (auto & [key, value] : element_obj) {
+      total_word_count += word_count(key);
+      total_word_count += json_word_count(&value);
+    }
+    return total_word_count;
+  } else if (input->is_number() || input->is_boolean()) {
+    // Booleans and numbers are constant
+    return 1;
+  }
```
yuval-k, line 442: do we want to add a debug assert / log if nothing matches?

EItanya: I don't think so, if you pass in some empty struct the result may be zero, but we already have logs to show the inputs and things so that should be enough to deduce.

yuval-k (approved):

nfuden:
🚀

PR #347, opened 2024-06-13
delete callback

EItanya:


source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -492,6 +495,13 @@ json TransformerInstance::base64url_decode_callback(const inja::Arguments &args)
   return Base64Url::decode(input);
 }
 
+json TransformerInstance::deletekey_callback(const inja::Arguments &args) {
+  auto json_input  = args.at(0)->get<json::object_t>();
+  const std::string & json_key = args.at(1)->get_ref<const std::string &>();
```
jbohanon, line 500: If we have `advanced_templates` set we will need to make a json pointer from the arg https://json.nlohmann.me/features/json_pointer/#introduction

PR #350, opened 2024-07-03
Add new `MergeJsonKeys` functionality to transformation filter.

EItanya:
Add a new feature to the `inja_transformer` called `MergeJsonKeys` which allows merging of rendered templates into the JSON body without needing to rewrite the entire body


source/extensions/filters/http/transformation/inja_transformer.h
```diff
@@ -154,6 +154,7 @@ class InjaTransformer : public Transformer {
 
   absl::optional<inja::Template> body_template_;
   bool merged_extractors_to_body_{};
+  std::vector<std::tuple<std::string, bool, inja::Template>> merge_templates_;
```
jbohanon, line 157: nit: can you add a comment about what these fields are so when cruising through the `.h` it's more clear?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -945,6 +960,38 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
   } else if (merged_extractors_to_body_) {
     std::string output = json_body.dump();
     maybe_body.emplace(output);
+  } else if (merge_templates_.size() > 0) {
```
jbohanon, line 963: nit: https://cplusplus.com/reference/vector/vector/empty/
```suggestion
  } else if (!merge_templates_.empty()) {
```

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -945,6 +960,38 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
...
+        name_to_split = name_to_split.substr(pos + 1);
+      }
+      const auto rendered = instance_->render(std::get<2>(merge_template));
+      // Do not overwrite with empty unless specified
+      if (rendered.size() > 0 || std::get<1>(merge_template)) {
+        try {
+          auto rendered_json = json::parse(rendered);
+          (*current)[std::string(name_to_split)] = rendered_json;
+        } catch (const std::exception &e) {
+          ASSERT("failed to parse merge_json_key output");
```
jbohanon, line 987: should we emit this to the stream log instead of using `ASSERT`?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -945,6 +960,38 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
...
     maybe_body.emplace(output);
+  } else if (merge_templates_.size() > 0) {
+
+    for (const auto &merge_template : merge_templates_) {
+      const std::string &name = std::get<0>(merge_template);
+      
+      // prepare variables for non-advanced_templates_ scenario
+      absl::string_view name_to_split;
+      json* current = nullptr;
+      // if (!advanced_templates_) {
```
jbohanon, line 971: If we don't end up supporting json pointer templates, I think we should keep this check in and throw in the constructor at config time

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: I wanted to, but this part of the code doesn't have access to those macros, I can look into adding them

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: I need to understand better what would be entailed, I honestly don't really know the different 🙈

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -236,13 +236,13 @@ TransformerInstance::TransformerInstance(ThreadLocal::Slot &tls, Envoy::Random::
   env_.add_callback("clusterMetadata", 1, [this](Arguments &args) {
     return cluster_metadata_callback_deprecated(args);
   });
-  env_.add_callback("cluster_metadata", 1, [this](Arguments &args) {
+  env_.add_callback("cluster_metadata", [this](Arguments &args) {
```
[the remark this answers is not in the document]
EItanya: Yea it is! Essentially it makes it variadric. These callbacks can take 1 or 2, they allow you to specify the metadata namespace if you pass it in the 2nd arg

source/extensions/filters/http/transformation/inja_transformer.cc
jbohanon: Ah yeah we would need the filter class to inherit the loggable a la https://github.com/solo-io/envoy-gloo/blob/main/source/extensions/filters/http/aws_lambda/aws_lambda_filter.h#L28C25-L28C61

source/extensions/filters/http/transformation/inja_transformer.cc
jbohanon: It's largely just handling of `/` and `~`

spec: https://datatracker.ietf.org/doc/html/rfc6901

our fork to re-introduce support in inja: https://github.com/solo-io/inja/pull/6

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -772,6 +772,21 @@ InjaTransformer::InjaTransformer(const TransformationTemplate &transformation,
     merged_extractors_to_body_ = true;
     break;
   }
+  case TransformationTemplate::kMergeJsonKeys: {
+    if (transformation.parse_body_behavior() == TransformationTemplate::DontParse) {
+      throw EnvoyException("MergeJsonKeys requires parsing the body");
```
nfuden, line 777: https://github.com/envoyproxy/envoy-mobile/issues/176
Any way we can also avoid an exception here and just return out?
Not a big deal if not as it seems to mostly be a build sizing thing rather than a true runtime performance thing

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -267,6 +267,11 @@ message TransformationTemplate {
     // If you want to nest elements inside the body, use dot separator in the
     // extractor name.
     MergeExtractorsToBody merge_extractors_to_body = 6;
+    // A set of key-value pairs to merge into the JSON body.
+    // Each value will be rendered separately, and then placed into the JSON body at
```
nfuden, line 271:
```suggestion
    // Each value will be rendered separately, and then placed into the JSON body at
    // Cannot be used on any other section other than the body
```

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -945,6 +960,38 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
...
+
+    for (const auto &merge_template : merge_templates_) {
+      const std::string &name = std::get<0>(merge_template);
+      
+      // prepare variables for non-advanced_templates_ scenario
+      absl::string_view name_to_split;
+      json* current = nullptr;
+      // if (!advanced_templates_) {
+      name_to_split = name;
+      current = &json_body;
```
nfuden, line 973: is the intent to mutate what is in   typed_tls_data.context_?
Or should this be a copy?

I think the answer is that it should mutate whats in the context_ field but I figure I should at least ask

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -759,6 +759,47 @@ TEST_F(InjaTransformerTest, UseBodyFunction) {
...
+  envoy::api::v2::filter::http::MergeJsonKeys_OverridableTemplate tmpl;
+  (*tmpl.mutable_tmpl()) = inja_template;
+  (*transformation.mutable_merge_json_keys()->mutable_json_keys())["ext2"] = tmpl;
+
+  InjaTransformer transformer(transformation, rng_, google::protobuf::BoolValue(), tls_);
+
+  NiceMock<Http::MockStreamDecoderFilterCallbacks> callbacks;
+  Buffer::OwnedImpl body("{\"ext1\":\"123\"}");
+  transformer.transform(headers, &headers, body, callbacks);
+  EXPECT_EQ(body.toString(), "{\"ext1\":\"123\",\"ext2\":\"/foo\"}");
```
nfuden, line 781: Lets add a test where the added fields specify an existing json field

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -759,6 +759,47 @@ TEST_F(InjaTransformerTest, UseBodyFunction) {
...
+}
+
+TEST_F(InjaTransformerTest, MergeJsonKeysEmpty ) {
```
nfuden, line 784: Can we add a test where the body is empty and one where its empty json?
Buffer::OwnedImpl body("");
Buffer::OwnedImpl body("{}");

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: This is a copy/paste of the logic for extractors, but that is a very very good point, any operations done on the body will be impacted by this. I think it's a very valid question of whether this should happen before or after the rest of the transformations. Extractors directly modify the JSON body beforehand, so maybe we should do the same?

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: What do you mean `just return out`? We specifically want to Let the user know this isn't supported, we NEED to parse the body for this functionality to work

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: Done

source/extensions/filters/http/transformation/inja_transformer.cc
nfuden: Im dumb I thought that this was on the request path. Disregard

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: I'm now throwing an exception at config time

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -772,6 +772,23 @@ InjaTransformer::InjaTransformer(const TransformationTemplate &transformation,
     merged_extractors_to_body_ = true;
     break;
   }
+  case TransformationTemplate::kMergeJsonKeys: {
+    if (transformation.parse_body_behavior() == TransformationTemplate::DontParse) {
```
yuval-k, line 776: should this be `!=ParseAsJson` incase we add more options in the future?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -772,6 +772,23 @@ InjaTransformer::InjaTransformer(const TransformationTemplate &transformation,
...
+      throw EnvoyException("MergeJsonKeys requires parsing the body");
+    } else if (transformation.advanced_templates()) {
```
yuval-k, line 778: add this to the proto docs?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -772,6 +772,23 @@ InjaTransformer::InjaTransformer(const TransformationTemplate &transformation,
...
+      throw EnvoyException("MergeJsonKeys is not supported with advanced templates");
+    } 
+    try {
+      for (const auto &named_extractor : transformation.merge_json_keys().json_keys()) {
+        merge_templates_.emplace_back(std::make_tuple(named_extractor.first, named_extractor.second.override_empty(), instance_->parse(named_extractor.second.tmpl().text())));
+      }
+
+    } catch (const std::exception &e) {
+      throw EnvoyException(
+          fmt::format("Failed to parse merge_body_key template {}", e.what()));
```
yuval-k, line 788: what's merge_body_key? did you want to interpolate the key?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -772,6 +772,23 @@ InjaTransformer::InjaTransformer(const TransformationTemplate &transformation,
...
     break;
   }
+  case TransformationTemplate::kMergeJsonKeys: {
+    if (transformation.parse_body_behavior() == TransformationTemplate::DontParse) {
+      throw EnvoyException("MergeJsonKeys requires parsing the body");
+    } else if (transformation.advanced_templates()) {
+      throw EnvoyException("MergeJsonKeys is not supported with advanced templates");
+    } 
+    try {
+      for (const auto &named_extractor : transformation.merge_json_keys().json_keys()) {
```
yuval-k, line 782: i believe you can do this in c++ 20:
```
for (const auto &[name , tmpl] : transformation.merge_json_keys().json_keys()) {
```
to directly get the key and value

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -945,6 +962,38 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
...
+        name_to_split = name_to_split.substr(pos + 1);
+      }
+      const auto rendered = instance_->render(std::get<2>(merge_template));
+      // Do not overwrite with empty unless specified
+      if (rendered.size() > 0 || std::get<1>(merge_template)) {
+        try {
+          auto rendered_json = json::parse(rendered);
+          (*current)[std::string(name_to_split)] = rendered_json;
+        } catch (const std::exception &e) {
+          ENVOY_STREAM_LOG(debug, "failed to parse merge_json_key output: {}", callbacks, e.what());
```
yuval-k, line 989: merge_json_key?

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -945,6 +962,38 @@ void InjaTransformer::transform(Http::RequestOrResponseHeaderMap &header_map,
...
+          (*current)[std::string(name_to_split)] = rendered;
```
yuval-k, line 990: I'm not a fan of during different logic in a way that's not predictable (i.e. if we had an error parsing the json). can this be explicit config? i.e. parse template as json or not?

source/extensions/filters/http/transformation/inja_transformer.cc
EItanya: Going to nest the exception to have access to the key

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -267,6 +267,13 @@ message TransformationTemplate {
...
+    // the specified key.
+    // There are a number of important caveats to using this feature:
+    // * This can only be used when the body is parsed as JSON.
+    // * This option does NOT work with advanced tempaltes currently
```
nfuden, line 275:
```suggestion
    // * This option does NOT work with advanced templates currently
```

nfuden:
lgtm pending yuval check

source/extensions/filters/http/transformation/inja_transformer.cc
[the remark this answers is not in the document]
yuval-k: we should validate in parse_metadata that we have at least 1 arg

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -759,6 +759,86 @@ TEST_F(InjaTransformerTest, UseBodyFunction) {
...
+
+
+  // TransformationTemplate transformation;
+  // transformation.mutable_body()->set_text("{\"ext2\": \"{{header(\":path\")}\" }");
+
+  envoy::api::v2::filter::http::InjaTemplate inja_template;
+  inja_template.set_text("\"{{header(\":path\")}}\"");
+  envoy::api::v2::filter::http::MergeJsonKeys_OverridableTemplate tmpl;
+  (*tmpl.mutable_tmpl()) = inja_template;
+  (*transformation.mutable_merge_json_keys()->mutable_json_keys())["ext2"] = tmpl;
```
yuval-k, line 774: please add tests for the dot logic. include edge cases to make sure we don't crash. i.e.
- "foo.bar"
- "foo..bar"
- ".foo."
- "..."
- ".foo
- "bar."

yuval-k: also empty string?

EItanya: Discussed offline, I am going to remove this ability for now as it's not necessary. In order to ensure we can potentially add it back later I will throw an exception at `config_time` if a `"."` is present in the key

yuval-k (approved):

PR #363, opened 2024-09-16
metadata-subpaths

EItanya:
Allow usage of subpaths in Metadata funcs within `inja` templates.

The `TransformationTemplate` API now has a `string_delimiter` field to enable swapping the delimiter if necessary given the contents of the metadata keys. By default it is set to `:`.

### DIFF
--- a/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
+++ b/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
@@ -314,6 +314,12 @@ message TransformationTemplate {
   // rendered string values to be escaped in order to make valid JSON/YAML strings
   bool escape_characters = 12;
 
+  // The character to use as a delimiter when accessing dynamic metadata values
+  // in templates. Defaults to ':' if unset.
+  // See https://www.envoyproxy.io/docs/envoy/v1.31.1/configuration/observability/access_log/usage.html
+  // for more information.
+  string string_delimiter = 14;
+
 }
 
 // Defines an [Inja template](https://github.com/pantor/inja) that will be
--- a/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
+++ b/api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
@@ -316,7 +316,7 @@ message TransformationTemplate {
 
   // The character to use as a delimiter when accessing dynamic metadata values
   // in templates. Defaults to ':' if unset.
-  // See https://www.envoyproxy.io/docs/envoy/v1.31.1/configuration/observability/access_log/usage.html
+  // See https://www.envoyproxy.io/docs/envoy/v1.31.1/configuration/observability/access_log/usage.html#config-access-log-format-dynamic-metadata
   // for more information.
   string string_delimiter = 14;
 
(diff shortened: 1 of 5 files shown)


api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -314,6 +314,12 @@ message TransformationTemplate {
   // rendered string values to be escaped in order to make valid JSON/YAML strings
   bool escape_characters = 12;
 
+  // The character to use as a delimiter when accessing dynamic metadata values
+  // in templates. Defaults to ':' if unset.
+  // See https://www.envoyproxy.io/docs/envoy/v1.31.1/configuration/observability/access_log/usage.html
```
ashishb-solo, line 319:
```suggestion
  // See https://www.envoyproxy.io/docs/envoy/v1.31.1/configuration/observability/access_log/usage.html#config-access-log-format-dynamic-metadata
```

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -314,6 +314,12 @@ message TransformationTemplate {
...
+  // for more information.
+  string string_delimiter = 14;
```
yuval-k, line 321: why is this configurable? can we just have it be an internal constant?

test/extensions/filters/http/transformation/inja_transformer_test.cc
```diff
@@ -1237,6 +1237,71 @@ TEST_F(InjaTransformerTest, ParseFromClusterMetadata) {
...
+      {SoloHttpFilterNames::get().Transformation,
+       struct_obj});
+  ON_CALL(callbacks.stream_info_, dynamicMetadata())
+      .WillByDefault(testing::ReturnRefOfCopy(meta));
+
+  Buffer::OwnedImpl body("1");
+  transformer.transform(headers, &headers, body, callbacks);
+  EXPECT_EQ(body.toString(), "[1,2]");
+}
+
```
ashishb-solo, line 1304: Is it possible to add a test where the delimiter is not the default value (ie. `':'`)?

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -314,6 +314,12 @@ message TransformationTemplate {
   // rendered string values to be escaped in order to make valid JSON/YAML strings
   bool escape_characters = 12;
 
+  // The character to use as a delimiter when accessing dynamic metadata values
+  // in templates. Defaults to ':' if unset.
```
yuval-k, line 318: why not default to dot? feels more natural?

EItanya: Because there are very often `.` in the keys, such as `envoy.filter.http` etc.

EItanya: Also, I wanted to mirror the access log formatting

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
EItanya: I don't have a strong feeling about this, can always expose it later

yuval-k (approved):

PR #371, opened 2024-10-16
bazel: 1.32

nfuden:
1.32


bazel/extensions/extensions_build_config.bzl
```diff
@@ -239,7 +247,8 @@ EXTENSIONS = {
 
     "envoy.resource_monitors.fixed_heap":               "//source/extensions/resource_monitors/fixed_heap:config",
     "envoy.resource_monitors.injected_resource":        "//source/extensions/resource_monitors/injected_resource:config",
-    "envoy.resource_monitors.downstream_connections":   "//source/extensions/resource_monitors/downstream_connections:config",
+    "envoy.resource_monitors.global_downstream_max_connections":   "//source/extensions/resource_monitors/downstream_connections:config",
```
andy-fong, line 250: Just want to make sure this is not a cut-and-paste error but intentionally changing the name but not the value?

andy-fong: I did a quick search and I think I answer my own question.

changelog/v1.32.1-patch1/update-to-upstream-envoy-v1.31.yaml
```diff
@@ -0,0 +1,10 @@
+changelog:
```
andy-fong, line 1: The file name has wrong version.

bazel/extensions/extensions_build_config.bzl
nfuden: Yep and we have a test that checks for these. So this would be an interesting change for us except we dont use it today in the control plane.

changelog/v1.32.1-patch1/update-to-upstream-envoy-v1.31.yaml
nfuden: updated!

andy-fong (approved):

PR #393, opened 2025-01-10
1.31: test/transforms: add failing test for overwritting extraction source

nfuden:
will forward port after

Change extraction and their storage mechanism (extractions) to use std::string rather than absl::string_view. 
This lets us avoid the bad usage of string_views per https://abseil.io/docs/cpp/guides/strings.

Previously this would ONLY affect configuration where a header was both extracted from and set in the same resolution.

See the breaking test https://github.com/solo-io/envoy-gloo/pull/395

https://storage.googleapis.com/solo-public-build-logs/logs.html?buildid=d0c7341e-a39c-4c18-9823-535dd36333d7


test/extensions/filters/http/transformation/transformation_filter_test.cc
```diff
@@ -525,6 +526,43 @@ TEST_F(TransformationFilterTest, StagedFilterRequestConfig) {
...
+            ":path": {text: "/{{extraction(\"ext1\")}}/somethingreallyreallylongsowecanevictmemoryorsomething"}
+            "x-foo": {text: "{{extraction(\"ext2\")}}"}
+  )EOF";
+  TestUtility::loadFromYaml(match_string, route_config_);
+
+  initFilter();
+  auto res = filter_->decodeHeaders(request_headers, true);
+  EXPECT_EQ(Http::FilterHeadersStatus::Continue, res);
+  EXPECT_EQ(request_headers.get_(":path"), "/abc/somethingreallyreallylongsowecanevictmemoryorsomething");
+  // enforces that we dont care that hte source has changed
```
EItanya, line 562:
```suggestion
  // enforces that we dont care that the source has changed
```

EItanya (commented):
LGTM, just one nit

EItanya (approved):

PR #396, opened 2025-01-21
downgrade control plane

nfuden:


go/go.mod
```diff
@@ -4,14 +4,15 @@ go 1.23.1
 
 require (
 	github.com/cncf/xds/go v0.0.0-20241223141626-cff3c89139a3
-	github.com/envoyproxy/go-control-plane/envoy v1.32.2
+	github.com/envoyproxy/go-control-plane v0.13.2
```
lgadban, line 7: i actually think this is too recent to match istio's; v0.13.2 is when the did the split: https://github.com/envoyproxy/go-control-plane/releases/tag/v0.13.2

istio is using a commit before this release, https://github.com/envoyproxy/go-control-plane/commit/fc612d4a3afabcbec1906fcd28df0f46adf6999b

yuval-k:
i think we can close this for now as it's not needed

PR #397, opened 2025-01-31
support data source in transform

yuval-k:
This PR introduces 2 new features for transformation.

First is the trim option which will trim ascii whitespace from both the start and end of the value provided.
This can be used like so: {{ trim(my_super_cool_value) }}

Second is the datasource option. This pulls from a mounted datasource to provide data or an empty string.
This can be used like so {{ data_source("my_cool_mounted_file") }}
One may often want to use the trim functionality in combination with this new feature.


source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -394,6 +400,28 @@ json TransformerInstance::extracted_callback(const inja::Arguments &args) const
...
+  const std::string &name = args.at(0)->get_ref<const std::string &>();
+  const auto value_it = ctx.data_sources_->find(name);
+  if (value_it != ctx.data_sources_->end()) {
+    return value_it->second->data();
+  }
+  return "";
+}
+
+json TransformerInstance::trim_callback(const inja::Arguments &args) const {
+  std::string s = args.at(0)->get_ref<const std::string &>();
```
nfuden, line 414: any reason to not use absl? https://github.com/abseil/abseil-cpp/blob/acf3390ca28edf1438fa896602ffede2a7dff103/absl/strings/ascii.h#L231

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -394,6 +400,28 @@ json TransformerInstance::extracted_callback(const inja::Arguments &args) const
...
 }
 
+json TransformerInstance::data_source_callback(const inja::Arguments &args) const {
+  const auto& ctx = tls_.getTyped<ThreadLocalTransformerContext>();
+  const std::string &name = args.at(0)->get_ref<const std::string &>();
+  const auto value_it = ctx.data_sources_->find(name);
+  if (value_it != ctx.data_sources_->end()) {
+    return value_it->second->data();
+  }
+  return "";
```
nfuden, line 410: Should we make this with a default return?

source/extensions/filters/http/transformation/inja_transformer.cc
yuval-k: didn't know absil had this!

source/extensions/filters/http/transformation/inja_transformer.cc
yuval-k: this return represents a user error. i.e. the user asking for a data source the user did not define. so i don't think we need a default here, as this should never happen.

source/extensions/filters/http/transformation/inja_transformer.cc
nfuden: So it looks like since c++ 17 it is safe to rely on std::isstring but upstream does use the ascii whitspace stripper so it does feel like a nicer strategy imo.
That being said i am fine either way.

nfuden (approved):

api/envoy/config/filter/http/transformation/v2/transformation_filter.proto
```diff
@@ -230,6 +231,13 @@ message TransformationTemplate {
   // "my-extractor" extractor.
   map<string, Extraction> extractors = 2;
 
+  // Data sources. use the `data_source`
+  // function to access data_sources in the template (e.g. '{{
+  // data_source("my_file") }}').
+  map<string, envoy.config.core.v3.DataSource> data_sources = 17;
+  // max size of data source. 4096 bytes by default.
+  uint32 max_size = 18;
```
EItanya, line 239: Can we name this `data_source_max_size` so it's more clear what it specifically applies to

source/extensions/filters/http/transformation/inja_transformer.cc
```diff
@@ -715,6 +738,23 @@ InjaTransformer::InjaTransformer(const TransformationTemplate &transformation,
...
     extractors_.emplace_back(std::make_pair(it->first, it->second));
   }
+  uint32_t max_size = 4096;
+  if (transformation.max_size() > 0) {
+    max_size = transformation.max_size();
+  }
+
+  const auto &data_sources = transformation.data_sources();
+  for (auto it = data_sources.begin(); it != data_sources.end(); it++) {
+        auto provider_or_error = Envoy::Config::DataSource::DataSourceProvider::create(
```
EItanya, line 748: I'm not super familiar with the `Envoy::Config::DataSource::DataSourceProvider::create`. Does this load the data on the data path or pre-load? Doing the former seems potentially expensive.

nfuden: https://github.com/envoyproxy/envoy/blob/release/v1.32/source/common/router/config_impl.cc#L612

yuval-k: yes, it watches for file updates

EItanya (approved):

PR #415, opened 2025-04-08
envoy 1.34 bump for istio pipeline

nfuden:
edit: merging this pull request will not fully address https://github.com/solo-io/envoy-gloo-ee/issues/909 - we still need to merge the `envoy-gloo-ee` pull request as well: https://github.com/solo-io/envoy-gloo-ee/pull/929


ashishb-solo:
/kick-ci ? not sure why buildbot isn't running

source/common/matcher/solo_matcher.cc
```diff
@@ -120,6 +120,9 @@ class CompiledStdMatcher : public Regex::CompiledMatcher {
 public:
   CompiledStdMatcher(std::regex &&regex) : regex_(std::move(regex)) {}
 
+  // TODO(nfuden) refactor matchers or find a way to get back onto upstream matchers
+  const std::string& pattern() const override { return EMPTY_STRING; }
```
andy-fong, line 124: Not sure what the implication of this is but this can be easily fixed because I only see one place uses this class: https://github.com/solo-io/envoy-gloo/blob/d545af2580ab71dd1dcd59ea36e89236c4456191/source/common/matcher/solo_matcher.cc#L153-L154
basically, accept a string & in the construction and call Solo::Regex::Utility::parseStdRegex(regex) inside this constructor, then we can save the original string to return for pattern().

bazel/repositories.bzl
```diff
@@ -67,11 +67,6 @@ def _repository_impl(name, **kwargs):
 
 def envoy_gloo_dependencies():
     _repository_impl("envoy", patches=[
-        "@envoy_gloo//bazel/foreign_cc:001-tlv-support.patch", #remove in  1.34
-        # revert upstream's cache filter from using a separate HttpAsyncClient
```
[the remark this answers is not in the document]
ashishb-solo: Good catch - see 948342c62d0819ed49406028fab1987d09340194 for the reversion

changelog/v1.34.0-patch00/bumpenvoy.yaml
```diff
@@ -0,0 +1,9 @@
+changelog:
+  - type: DEPENDENCY_BUMP
+    issueLink: https://github.com/solo-io/envoy-gloo-ee/issues/887
+    dependencyOwner: envoyproxy
+    dependencyRepo: envoy
+    dependencyTag: 1.34.0
+    resolvesIssue: false
+    description: >-
+      Bumps to a prerelease of envoy 1.34.0 to allow for usage of dynamic modules
```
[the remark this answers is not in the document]
ashishb-solo: Will pull this in ae765ac6c2a5cfd3e6a40fa51f52e1e8f8e634dc

source/common/matcher/solo_matcher.cc
```diff
@@ -118,10 +118,12 @@ class PathMatcherImpl : public BaseMatcherImpl {
 
 class CompiledStdMatcher : public Regex::CompiledMatcher {
 public:
-  CompiledStdMatcher(std::regex &&regex) : regex_(std::move(regex)) {}
+  CompiledStdMatcher(const std::string &pattern) :
```
andy-fong, line 121: The better pattern for this is:
```suggestion
  CompiledStdMatcher(const std::string pattern) :
  pattern_(std::move(pattern)),
```
This way, the caller can actually move the string into here.

source/common/matcher/solo_matcher.cc
```diff
@@ -144,14 +146,14 @@ class CompiledStdMatcher : public Regex::CompiledMatcher {
   }
 
 private:
+  const std::string &pattern_;
```
andy-fong, line 149: This cannot be a reference as the reference can go out of scope. Need to be a string.

andy-fong (changes requested):
In terms of memory used to hope the string. Unless we have millions of this object and they tend to do long regex pattern, I don't think it should be a concern. I looked all the similar class in envoy using re2 regex, they do the same of storing the string in the object as well.

source/common/matcher/solo_matcher.cc
ashishb-solo: Should be fixed in 31208c53037892a0e894a0d91573d7719f474bb5 but would appreciate if you could take a second look. I'm still a bit shaky on exactly how move semantics work.

I also thought about using `std::move` when calling the `CompiledStdMatcher` constructor too:

```
diff --git a/source/common/matcher/solo_matcher.cc b/source/common/matcher/solo_matcher.cc
index 48d27b6..7165efe 100644
--- a/source/common/matcher/solo_matcher.cc
+++ b/source/common/matcher/solo_matcher.cc
@@ -153,7 +153,7 @@ private:
 class StdRegexEngine : public Regex::Engine {
 public:
   absl::StatusOr<Regex::CompiledMatcherPtr> matcher(const std::string &regex_pattern) const override {
-    return std::make_unique<CompiledStdMatcher>(std::move(regex_pattern));
+    return std::make_unique<CompiledStdMatcher>(regex_pattern);
   }
 };
```

but then i reverted it because i wasn't sure if the calling function was still using the string, so i thought it might be unsafe. let me know what you think - thanks!

source/common/matcher/solo_matcher.cc
ashishb-solo: let's mark this as resolved and move the conversation to https://github.com/solo-io/envoy-gloo/pull/415#discussion_r2077773869

source/common/matcher/solo_matcher.cc
andy-fong: correct, you cannot use move here because it's a reference, you don't know where that string is and if the user still needs it. Like my suggestion for the CompiledStdMatcher, if you want to use move here, you can just change that regex_pattern to be pass by value, then it's save to move. You are making a copy anyway, the key is to make a copy only once, whether the compiler does it or you manually do it. When the compiler does it, it can optimize where it can.

andy-fong: wait, you cannot change the function signature here because is "override", so what you have now is correct.

PR #416, opened 2025-04-10
Migrate AI Transforamtion from generic template to native C++

andy-fong:
Originally, this was done in envoy-gloo-ee's TransformationEE filter but we have decided to move to envoy-gloo for open-source.

## Background:
The primary goal of this change is to move the generic inja template transformation we set for AI Gateway to native C++ code so it's more flexible to support features like Unified API, Multi-Provider (different schema or platform) failover/retries. 
For more details, refer to this [EP](https://github.com/solo-io/solo-projects/blob/fd133414ac6b02f9db4c9f4c53e2bcce20d23e92/design-docs/ai-unified-api/7144.md)

There is no functional change nor user facing API changes (might add some new field in gloo's ai.proto but it won't break any user) 

## Here is the features that are migrated from generic inja template to native c++:

- path transformation
- auth token header injection
- model modification in the body
- field defaults
- prompt enrichment

## Some bug fixes and behavior changes:
- use `x-goog-api-key` key header for gemini key instead of putting it in the qs param. See [comment here](https://github.com/solo-io/envoy-gloo-ee/pull/910/files#diff-0e6cae80a9f7e9e1ca753c5c0165efcacd934588da6d75657ac021da827bf097R476).
- Now prompt enrichment will work correct on all providers (even Anthropic that uses a single `system` field and Gemini uses a different `system_instruction` field for system prompt).
- In auth passthru mode, if both the key header and authorization header exists, we will not touch the key header. Previously, we would still extract the token from the authorization header and override the existing key header.
- Now OpenAI platform API endpoints like embeddings, file, images ....etc will be passed thru.
- If we allow different providers that use the same json_schema from the control plane, now the path will be translated correctly when failover (eg supporting multipool with OpenAI and Azure OpenAI)
- previously we ignore the `routeType: CHAT_STREAMING` except it's Gemini and Vertex as we construct the URL for them. Now this will also turn on the streaming feature in OpenAI and Anthropic by setting `"stream": true` in the body. For OpenAI, it will also set the `include_usage` field to true in the `stream_options` field. Setting the RouteType to CHAT will not explicitly turn off streaming if the original OpenAI request body has streaming turned on. It will just leave it as is. 

Required by:
- gloo PR: https://github.com/solo-io/gloo/pull/10752
- solo-projects: https://github.com/solo-io/solo-projects/pull/8135


source/extensions/filters/http/transformation/ai_transformer.cc
```diff
@@ -0,0 +1,635 @@
...
+ */
+std::string replaceModelInPath(std::string_view original_path, std::string_view model) {
+  return absl::StrReplaceAll(original_path, {{"{{model}}", model}});
+}
+
+/**
+ * @brief lazily constructed singleton regex to match OpenAI Platform API endpoints
+ *        This same regex is also used in the AI extproc in solo-projects to bypass OpenAI Platform API from going to 
+ *        through the extproc, so if this regex is change, it should be changed in 
+ *        solo-projects/projects/ai-extension/ai_extension/ext_proc/server.py (search for open_ai_platform_api_regex)
```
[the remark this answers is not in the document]
andy-fong: We have not port this to kgateway yet, when we do, I will make sure to add the comment here as well.

test/extensions/filters/http/transformation/ai_transformer_test.cc
```diff
@@ -0,0 +1,948 @@
...
+      return "";
+    }
+    return headers_.Path()->value().getStringView();
+  }
+
+  void setPath(std::string_view path) {
+    headers_.setPath(path);
+  }
+
+  std::string_view getHeaderValue(std::string_view key) {
```
npolshakova, line 283: If the header is already set, does the filter overwrite it? For example, do we want the `api-key` to get overwritten if it's already present?

test/extensions/filters/http/transformation/ai_transformer_test.cc
```diff
@@ -0,0 +1,948 @@
...
+
+using json = nlohmann::json;
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace Transformation {
+
+namespace {
+
+using json = nlohmann::json;
```
npolshakova, line 31: Is this repeated on line 23?

source/extensions/filters/http/transformation/transformation_factory.cc
```diff
@@ -29,6 +30,9 @@ TransformerConstSharedPtr Transformation::getTransformer(
     auto config = Config::Utility::translateAnyToFactoryConfig(transformation.transformer_config().typed_config(), context.messageValidationContext().staticValidationVisitor(), factory);
     return factory.createTransformer(*config, transformation.log_request_response_info(), context);
   }
+  case envoy::api::v2::filter::http::Transformation::kAiTransformation:
```
npolshakova, line 33: Is this a typo? `kAiTransformation`?

source/extensions/filters/http/transformation/ai_transformer.cc
```diff
@@ -0,0 +1,635 @@
...
+  }
+
+  std::vector<std::string> elements = absl::StrSplit(key, delimiter);
+  const ProtobufWkt::Value &value = Envoy::Config::Metadata::metadataValue(
+      endpoint_metadata.get(), filter_name, elements);
+
+  switch (value.kind_case()) {
+  case ProtobufWkt::Value::kStringValue: {
+    return value.string_value();
+    break;
```
npolshakova, line 49: [nit] Do you need the break after the returns in this file?

source/extensions/filters/http/transformation/ai_transformer.cc
```diff
@@ -0,0 +1,635 @@
...
+    field_defaults_.emplace_back(field_default);
+  }
+};
+
+std::tuple<bool, bool> AiTransformer::transformHeaders(Http::RequestHeaderMap *request_map,
+                                  Envoy::Upstream::MetadataConstSharedPtr endpoint_metadata,
+                                  Http::StreamFilterCallbacks &callbacks,
+                                  const std::string &model) const {
+  std::string path;
+  bool in_bypass_mode = false;
```
npolshakova, line 427: Is `in_bypass_mode` for skipping transformations and passed to the provider as-is?

source/extensions/filters/http/transformation/ai_transformer.cc
```diff
@@ -0,0 +1,635 @@
...
+        AiTransformerConstants::get().GEMINI_STREAM_QS_PARAM);
+    } else {
+      absl::StrAppend(&path, AiTransformerConstants::get().GEMINI_GENERATE_CONTENT);
+    }
+
+    // Gemini doc is still using the `key` qs param but the Google GenAI sdk has switched to use the `x-goog-api-key` header.
+    // Here is the reason we also switch to the header:
+    //       1) currently the auth passthrough only work because the gemini sdk is using this header.
+    //          If someone is using the `key` qs param, passthrough never works. (Should we support that use case?)
+    //       2) putting the key in the qs param will probably get logged in access log which is a security concern
```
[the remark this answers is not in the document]
npolshakova: Does that mean we won't support Gemini token passthrough (I don't think we currently do either)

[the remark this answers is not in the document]
andy-fong: This new code will still support Gemini token passthrough when the original request uses the x-goog-api-key header including converting token from `authorization` header into this key header. Previously, it just happens to "work" because the header got pass through like any other headers. Gemini token passthrough NEVER works if the original request uses qs param because we drop the entire URL and rewrite with our own. It's easy to make it work here if we want to.

source/extensions/filters/http/transformation/transformation_factory.cc
andy-fong: no, that's how it's generated from protobuf.

test/extensions/filters/http/transformation/ai_transformer_test.cc
andy-fong: it will ONLY be overwritten if we are not in auth pass through mode but this function is just a helper to get the value and it does treat non-exist and empty value the same.

test/extensions/filters/http/transformation/ai_transformer_test.cc
andy-fong: oops, will remove.

source/extensions/filters/http/transformation/ai_transformer.cc
andy-fong: `in_bypass_mode` bypass any body and path transformation but will still inject auth tokens.

source/extensions/filters/http/transformation/transformation_factory.cc
nfuden: doesnt the k mean constant or something

source/extensions/filters/http/transformation/ai_transformer.cc
```diff
@@ -0,0 +1,635 @@
...
+    ENVOY_STREAM_LOG(warn, "endpoint_metadata is null.", callbacks)
+    return;
+  }
+
+  auto model = lookupEndpointMetadata(endpoint_metadata, "model");
+  auto [in_bypass_mode, update_model_in_body] = transformHeaders(request_map, endpoint_metadata, callbacks, model);
+  if (in_bypass_mode || body.length() == 0) {
+    return;
+  }
+
```
nfuden, line 627: do we need safeguards against large body sizes?

source/extensions/filters/http/transformation/ai_transformer.cc
```diff
@@ -0,0 +1,635 @@
...
+      }
+      setProviderKeyHeader(request_map, AiTransformerConstants::get().AnthropicApiKeyHeader, auth_token, in_auth_token_passthru_mode);
+    } else {
+      setBearerAuthTokenHeader(request_map, auth_token, in_auth_token_passthru_mode);
+    }
+  }
+
+  if (!path.empty()) {
+    ENVOY_STREAM_LOG(debug, "changing path from {} to {}", callbacks, original_path, path);
+    request_map->setPath(path);
```
nfuden, line 516: Do we need to make sure that this only happens along with clear_route_cache?

source/extensions/filters/http/transformation/ai_transformer.cc
andy-fong: Because this is part of the Transformation filter, it also has a body size limit:
https://github.com/solo-io/envoy-gloo/blob/42d799a0630dcc58ca5acc4566cd8118d78f95bf/source/extensions/filters/http/transformation/transformation_filter.cc#L70
which I think is base on a envoy config setting.

source/extensions/filters/http/transformation/ai_transformer.cc
andy-fong: We setup this up as an upstream filter, I am not sure if clear_route_cache works for upstream filter but I don't think we need that for this path transformation. Maybe @EItanya can chime in here.

source/extensions/filters/http/transformation/ai_transformer.cc
```diff
@@ -0,0 +1,635 @@
...
+    std::string output;
+    auto status = ProtobufUtil::MessageToJsonString(value.list_value(), &output);
+    return output;
+    break;
+  }
+  default: {
+    break;
+  }
+  }
+  return "";
```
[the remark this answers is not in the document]
andy-fong: I cut and paste these from AI generated code .... lol

source/extensions/filters/http/transformation/ai_transformer.cc
```diff
@@ -0,0 +1,635 @@
...
+
+  std::string_view original_path = getRequestPath(request_map);
+  if (provider == AiTransformerConstants::get().PROVIDER_AZURE) {
+    if (model.empty()) {
+      ENVOY_STREAM_LOG(warn, "Azure OpenAI: required model setting is missing!", callbacks);
+    }
+    path = replaceModelInPath(lookupEndpointMetadata(endpoint_metadata, "path"), model);
+    setProviderKeyHeader(request_map, AiTransformerConstants::get().AzureApiKeyHeader, auth_token, in_auth_token_passthru_mode);
+  } else if (provider == AiTransformerConstants::get().PROVIDER_GEMINI) {
+    if (model.empty()) {
```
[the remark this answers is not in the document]
nfuden: its kinda problematic in its own right

npolshakova (approved):

source/extensions/filters/http/transformation/ai_transformer.cc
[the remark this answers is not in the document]
EItanya: Envoy has `RELEASE_ASSERT` or something like that. Where it won't crash in prod.

source/extensions/filters/http/transformation/ai_transformer.cc
EItanya: It does not, routing is already don't in the upstream chain.

source/extensions/filters/http/transformation/ai_transformer.cc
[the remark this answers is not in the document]
andy-fong: Changed this to ASSERT in [4bf457a08a9dd3d4d8b60ded659f9a3e7d9db65f](https://github.com/solo-io/envoy-gloo/pull/416/commits/4bf457a08a9dd3d4d8b60ded659f9a3e7d9db65f)

EItanya (approved):

PR #429, opened 2025-05-02
upgrade envoy to 1.31.8

andy-fong:


andy-fong:
/skip-changelog because we jump from 1.31.6 envoy to 1.31.8 envoy directly, changelog-bot is not happy.

sam-heilbron (approved):

PR #430, opened 2025-05-02
upgrade envoy to v1.32.6

andy-fong:


andy-fong:
/skip-changelog because we jump from 1.32.3 to 1.32.6 envoy directly and changelog-bot is not happy

sam-heilbron (approved):

PR #435, opened 2025-05-07
Avoid storing the original regex string twice

andy-fong:


source/common/matcher/solo_matcher.cc
```diff
@@ -182,7 +181,7 @@ class RegexMatcherImpl : public BaseMatcherImpl {
       absl::string_view path_view = path.getStringView();
       path_view.remove_suffix(query_string.length());
       if (regex_->match(path_view)) {
-        ENVOY_LOG(debug, "Regex requirement '{}' matched.", regex_str_);
+        ENVOY_LOG(debug, "Regex requirement '{}' matched.", regex_.pattern());
```
ashishb-solo, line 184: i think it needs to be this

```suggestion
        ENVOY_LOG(debug, "Regex requirement '{}' matched.", regex_->pattern());
```

ashishb-solo (approved):

PR #444, opened 2025-06-27
[AI] bedrock custom path override support

andy-fong:
Add ability to override the path for bedrock upstream. 
If there is a fullpath override, the control plane will set the "path" metadata. If not, it will set the "base_path" metadata.

### DIFF
--- a/source/extensions/filters/http/transformation/ai_transformer.cc
+++ b/source/extensions/filters/http/transformation/ai_transformer.cc
@@ -607,14 +607,16 @@ std::tuple<bool, bool> AiTransformer::transformHeaders(
 
   } else if (provider == AiTransformerConstants::get().PROVIDER_BEDROCK) {
     ASSERT(!model.empty(), "Bedrock: required model setting is missing!");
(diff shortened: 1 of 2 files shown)


source/extensions/filters/http/transformation/ai_transformer.cc
```diff
@@ -606,32 +609,37 @@ std::tuple<bool, bool> AiTransformer::transformHeaders(
...
+        if (enable_chat_streaming_) {
+          absl::StrAppend(
+              &path, AiTransformerConstants::get().GEMINI_STREAM_GENERATE_CONTENT,
+              "?", AiTransformerConstants::get().GEMINI_STREAM_QS_PARAM);
+        } else {
+          absl::StrAppend(&path,
+                          AiTransformerConstants::get().GEMINI_GENERATE_CONTENT);
+        }
       } else {
-        absl::StrAppend(&path,
```
andy-fong, line 629: All these changes are just checking "path" metadata is set first but the indentation change make this look more than it is changed.

PR #446, opened 2025-07-10
update to 1.35

yuval-k:
fix code breakages:
- aws default cred provider: https://github.com/envoyproxy/envoy/pull/39616
- matcher refactor: https://github.com/envoyproxy/envoy/pull/39673
patch 004 was merged upstream so can be removed


.bazelrc
andy-fong, line 1: With this change @yuval did, technically we don't need to maintain upstream.bazelrc anymore and I checked, I don't think we did any override on this repo. So, I am not changing this to use import like solo-projects.

PR #493, opened 2026-03-10
[1.34] upgrade envoy to v1.34.13

andy-fong:


.bazelrc
```diff
@@ -1,9 +1,5 @@
 # Envoy specific Bazel build/test options.
 
-# Use upstream's .bazelrc as a baseline, then override here where needed.
```
andy-fong, line 3: This file is basically a copy of upstream bazelrc file already, really no point to import and then duplicate everything. These have been removed in all other branches before but probably missed this one.

dhawton (approved):

PR #498, opened 2026-06-04
upgrade envoy to v1.36.7

andy-fong:


puertomontt (approved):

bazel/repositories.bzl
```diff
@@ -66,7 +66,9 @@ def _repository_impl(name, **kwargs):
             )
 
 def envoy_gloo_dependencies():
-    _repository_impl("envoy", patches=[
+    _repository_impl("envoy", 
+        patch_args = ["-p1"],
```
nfuden, line 70: Oh boo p1 :( XD

andy-fong: I had no preference on this but seems that everyone else is generating patches with prefix, istio, envoy ...etc so this is to be consistent so when copying patches around, people just use their normal workflow instead of finding out later the patch doesn't apply. 

We should chat about this when we meet in person on why p1 is bad ...lol.

nfuden: It was purely due to problems when double wrapping that would take place in enterprise

andy-fong: You mean the patch of patch which I called inception style patch. Glad they are no more.

